PR #1, opened 2018-07-18
Feature/profile information

Davlis:
Script to retrieve profile information from within Slack

Added command:

C1. ```profile @someone <fieldname>```

In ```someone``` user should put user nickname

In ```<fieldname>``` user should put one of available options: 

- real_name
- title
- phone
- skype
- email

Keep in mind that user also can query for custom fields


scripts/example.coffee
```diff
@@ -10,8 +10,8 @@
 
 module.exports = (robot) ->
 
-  robot.hear //i, (res) ->
-    res.send "Badgers? BADGERS? WE DON'T NEED NO STINKIN BADGERS"
+  # robot.hear /badger/i, (res) ->
```
siepet: Maybe delete comments if we dont need them?

lubieniebieski: yes, delete!

lubieniebieski:
hmmm, just a general question, shouldn't we rather go for a modern ecmascript instead of coffescript?

scripts/profile.coffee
```diff
@@ -0,0 +1,38 @@
+module.exports = (robot) ->
+
```
lubieniebieski: it would be cool to start this file with some description of the plugin and listing of the commands available. Bonus points for some kind of documentation which we can parse automatically and build the always up-to-date list of commands in our Readme for instance. What do you think?

Davlis:
@lubieniebieski It was written in coffee, because template provided by `yo` was originally in coffee, also Hubot is documented for coffee, so I decided to not switch for ecmascript. I think that is a matter of taste.

btw it would be easy & fast to change to ecma at this moment

FieryCod:
@Davlis I will highly appreciate if you could please change the `.coffee` to `.js` 👍 That would be awesome :) I do not want to say that `coffeescript` is bad. Coffescript is a great tool and it did very well for couple of years but the case is that there are no so many coffescript developers here. But there are numerous of Node.js/Javascript developers who might help you :)

tools/docs-parser/utils.js
```diff
@@ -0,0 +1,57 @@
...
+function readDir(dir) {
+    return fs.readdirSync(dir);
+}
+
+function readFile(path) {
+    return fs.readFileSync(path, { encoding: 'utf8' });
+}
+
+function readFiles(dirpath) {
+    const filenames = readDir(dirpath);
```
FieryCod: Please consider using simple map fn

scripts/profile.js
```diff
@@ -0,0 +1,27 @@
...
+ module.exports = (robot) => {
+    robot.respond(/profile (.*) (.*)/i, (res) => {
+        const userMention = res.message.mentions.find(mention => mention.type === 'user');
+
+        if (!userMention) {
+            return res.send('Failed to parse user mention');
+        }
+
+        const askedField = res.match[2].toLowerCase();
+        const avaiableFields = ['real_name', 'title', 'phone', 'skype', 'email'];
```
lubieniebieski: is it possible to load this dynamically? Different orgs - different fields. It would be cool if that would be possible.

scripts/profile.js
```diff
@@ -0,0 +1,27 @@
+// Description:
+//  Script to retrieve profile information about mentioned user.
+//  User is allowed to ask for: real_name, title, phone, skype, email
+// 
+// Commands:
+//  bob profile <user> <field> - Send <user>'s <field> information
```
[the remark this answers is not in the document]
Davlis: @lubieniebieski I looked through some Hubot scripts and I didn't find any script with Slack-ish syntax documentation, also if we would use slack syntax in docs then that would affect our `README.md` documentation, so I don't think it's good idea.

If we really want to have some style for commands docs while typing `bob help` then we must change script that is responsible for listing commands, so that would be another fork of libary

scripts/profile.js
lubieniebieski: The second looks better:) I agree with you!

scripts/profile.js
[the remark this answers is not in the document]
lubieniebieski: we can add it later, no problem. I just like nice-looking things:)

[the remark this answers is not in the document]
Davlis: @lubieniebieski 
Then I think that we should fork lib responsible for `bob help` command, and modify it in proper way:) It definitely deserves separate pull request

PR #8, opened 2018-08-01
Structure proposal

Davlis:
Do not merge - this is only proposal

Generally, you can see that everything in `/scripts` directory were moved to `/scripts/src/scriptnamedir/*`, and on `/scripts` directory now you can see that we have only files that are responsible for adding services (e.g `SomeService`) to script and invoking it.

If we would have that structure, we could easily mock services that are used by scripts by adding them to `robot.brain` in test cases

What do you think about this solution?


scripts/src/profile/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+        const userMention = res.message.mentions.find(mention => mention.type === 'user');
+
+        if (!userMention) {
+            return res.send('Failed to parse user mention');
+        }
+
+        const { id, info: { slack: { profile } } } = userMention;
+
+        const askedField = res.match[2].toLowerCase();
```
pawdro: suggest moving `res.match[2]` to some variable with explanation

PR #14, opened 2018-08-06
[Feature] Custom Profile Fields

Davlis:
**Task description**

- Add `profile.config.json` file that allows user to map slack profile fields for custom names
- Add `profile.service.js` that is responsible for communicating with Slack API and fetching user profile with custom profile fields
- Move low level logic in `profile` script to `profile.service`
- Add proper response in `profile` script that informs user when asked fields is not defined in workspace
- Change to User factory in `/test/factories/UserFactory` to be consistent with Slack API responses
- Add new tests about custom profile fields (e.g `Location`, `Leader`)
- Introduce airbnb code style


FieryCod:
Love the way you write commit messages <3

src/profile/profile.service.js
```diff
@@ -0,0 +1,87 @@
...
+    return profile;
+  }
+
+  transformProfile(profile) {
+    const fields = profile.fields || {};
+
+    const customValues = Object.values(fields)
+      .map(field => field.label)
+      .reduce((acc, curr, index) => {
+        acc[curr] = Object.values(fields)[index].value;
```
FieryCod: Please consider removing .map then:
```
.reduce((acc, field, index) => ({
      ...acc,
      acc[field.label]: restblabla
  }), {})
```

Davlis:
@FieryCod Changes applied 🌖

pawdro:
A description is highly desirable.

PR #18, opened 2018-08-16
Add messages rendering engine

krzyurb:
**Functionality Description**

Implementation of messages renderer using [mustache.js](https://github.com/janl/mustache.js/) template system.

**Why?**

To move code from bot scripts, that respond to the user with complex messages, to the separated template file (`.message`) to make them easier lookup, debug and test.

**Usage example** 

Inside `index.js` of our bot script:

```
robot.respond(/summary (.*)/i, async (res) => {
  ...
  const results = ExampleService.getResults(res);
  const message = new MessageRenderer('summary', results);
  res.send(messagae);
}
```

Then we create `summary.message` inside `/templates` directory using 
`mustache.js` syntax.

```
This is an example of some important summary {{id}} for {{user}}.
{{city}} - {{date}}
```


utils/messageRenderer.js
```diff
@@ -0,0 +1,19 @@
+const Mustache = require('mustache');
```
Davlis: Please consider renaming this file to `MessageRenderer.js`

Davlis: Please consider to keep structure of this utility class to be the same as `RespondMultiplexer` then we could create some indexing file in `utils` directory and simply export our utility classes.

Example indexing file
```javascript
const MessageRenderer = require('./MessageRenderer');
const RespondMultiplexer = require('./RespondMultiplexer');

module.exports = {
  MessageRenderer,
  RespondMultiplexer,
};
```

Example import:
```javascript
const { MessageRenderer } = require('../utils');
```

Please note that it would make easier for us to import multiple utility classes into one module

Example below
```javascript
const { MessageRenderer, RespondMultiplexer } = require('../utils');
```

Davlis:
Great idea about introducing such class to our project 👍 
I left some comments for consideration

utils/messageRenderer.js
krzyurb: You're right

krzyurb: Nice! Now i get idea of that structure :)

krzyurb:
@Davlis Fixes done

PR #23, opened 2018-08-21
[Feature] Props

Getriax:
**Task description**

Implementation of props api, now bob should be able to give props to mentioned users.

give_props <users> <text>

**Examples**
give_props @some_person some text
give_props @some_person @another_person some text

Bob needs the **PROPS_API_TOKEN** environment variable to be able to perform that operation

Jira
TECH-90


scripts/props.js
```diff
@@ -0,0 +1,22 @@
+// Description:
+//  If you are authenticated it allows you to send props
+//
+//
+// Commands:
+//  bob give_props <user/s> <text> - give user/s the props with text
+//
+// Configuration:
+// PROPS_TOKEN: AUTH_TOKEN_TO_PROPS_APP
```
Davlis: Please consider writing something about `PROPS_TOKEN` - what is it? is it required? etc

src/props/api.js
```diff
@@ -0,0 +1,65 @@
...
+  getAllUsersFromProps() {
+
+    return new Promise((resolve, reject) => {
+      this.robot.http(PROPS_URL)
+        .header('Accept', 'application/json')
+        .path(GET_ALL_USERS_API_PATH)
+        .get()((err, resp, body) => {
+          if (err)
+            return reject(err);
+          if(resp.statusCode !== 200)
```
Davlis: Please try to follow `eslint` rules.

src/props/index.js
```diff
@@ -0,0 +1,74 @@
+const RespondMultiplexer = require('../../utils/RespondMultiplexer');
+const { MessageRenderer } = require('./../../utils');
```
Davlis: Consider writing 
```javascript
const { RespondMultiplexer, MessageRenderer } = require('../../utils');
```

src/props/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+module.exports = robot => {
+  const propsManager = async (res, match) => {
+    const { propsService } = robot.brain;
+
+    const propserSlackId = propsService.getUserSlackId(res);
+    if (!propserSlackId)
+      return res.send('It seems like you don\'t have an id');
```
Davlis: Please consider returning more descriptive information about missing id

src/props/api.js
```diff
@@ -0,0 +1,65 @@
...
+      throw new Error('GET DATA ERROR! ' + err);
+    });
+  }
+
+  giveProp(propserId, userIds, body, res) {
+    const data = JSON.stringify({
+      api_key: PROPS_API_TOKEN,
+      propser_id: propserId,
+      user_ids: userIds,
+      body
```
Davlis: Please follow eslint rules (trailing-comma)

If you are using `Visual Studio Code` editor please consider installing this https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
+class PropsService {
+  constructor(robot) {
+    this.robot = robot;
+    this.api = robot.brain.propsAPI;
+    // SLACK_UID:PROPS_ID
```
Davlis: Please consider removing this comment 😅 

Commenting in code are usually bad choice - we should try at best to provide more more descriptive function names / variables / etc instead of commenting

test/factories/PropsFactory.js
```diff
@@ -0,0 +1,32 @@
...
+const slackUser = (uid) => {
+  const id = uid || faker.random.uuid();
+  return {
+    id,
+    type: faker.random.word(),
+    info:
+      {
+       id,
+       name: faker.name.firstName() + '.' + faker.name.lastName(),
+       real_name: faker.name.firstName() + faker.name.lastName(),
```
Davlis: `name` and `real_name` would be different here

test/props.spec.js
```diff
@@ -0,0 +1,154 @@
...
+const PropsService = require('../src/props/props.service');
+const API = require('../src/props/api');
+
+const helper = new Helper('../src/props/index.js');
+
+const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+
+
+describe('give_props <users> <text> | tests', () => {
```
Davlis: Please consider removing `| tests` text, I believe that everyone will know that 😅

test/props.spec.js
```diff
@@ -0,0 +1,154 @@
...
+
+const PropsFactory = require('./factories/PropsFactory');
+const PropsService = require('../src/props/props.service');
+const API = require('../src/props/api');
+
+const helper = new Helper('../src/props/index.js');
+
+const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+
```
Davlis: Too many new lines here, please consider removing

test/props.spec.js
```diff
@@ -0,0 +1,154 @@
...
+
+describe('give_props <users> <text> | tests', () => {
+
```
Davlis: Please consider adding more test cases 

Here you can see some uncovered lines

screen shot 2018-08-21 at 17 53 04

test/props.spec.js
```diff
@@ -0,0 +1,154 @@
...
+  let room = null;
+  let propsService;
+  let propsAPI;
+
+  beforeEach(async () => {
+    room = helper.createRoom();
+    room.robot.brain.propsAPI = propsAPI = new API(null);
```
Davlis: Consider to not pass null values to class constructors

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
...
+    this.propsUsers;
```
Davlis: I believe you never use `this.propsUsers`

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
...
+  async giveProp(propserId, userIds, body) {
+   return await this.api.giveProp(propserId, userIds, body);
+  }
+
+  getUnrecognisableUsers(users) {
+    return users.filter(user => !this.userIsRecognisable(user.id));
+  }
+
+  userIsRecognisable(slackId) {
+    return slackId in this.propsUsers;
```
Davlis: `this.propsUsers` isn't always null here?

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  async getPropsUsers() {
+    return this.propsUsers || await this.getUsersFromServer();
```
Davlis: If you really want to do this in this way, please consider removing `await` here - it's unnecessary

**(super important)** You only fetch users from server **once**, what would happen when `Bob` server would run for few months, and we would have some new users in Slack workspace?

Can we fetch only one person from PROPS API? If yes then please consider changing to make a request every time `Bob` is called to call a props

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
...
+    .then(data => this.propsUsers = data)
+  }
+
+
+  getUserSlackId(res) {
+    return res.message.user.id;
+  }
+
+  getUsersFromMentions(res) {
+    return res.message.mentions
```
Davlis: Please consider making one liner here

src/props/api.js
```diff
@@ -0,0 +1,65 @@
+const PROPS_URL = 'https://netguru.props.devguru.co';
+const PROPS_API_PATH = '/api/v1';
+const PROPS_API_TOKEN = process.env.PROPS_API_TOKEN;
+const GET_ALL_USERS_API_PATH = `${PROPS_API_PATH}/users?api_key=${PROPS_API_TOKEN}`;
+const GIVE_PROPS_API_PATH = `${PROPS_API_PATH}/props`;
+const qs = require('querystring');
```
Davlis: We never use `querystring`, please consider removing it

src/props/api.js
```diff
@@ -0,0 +1,65 @@
...
+    return new Promise((resolve, reject) => {
+      this.robot.http(PROPS_URL)
+        .header('Accept', 'application/json')
+        .path(GET_ALL_USERS_API_PATH)
+        .get()((err, resp, body) => {
+          if (err)
+            return reject(err);
+          if(resp.statusCode !== 200)
+            return reject(resp.statusMessage);
+          const data = JSON.parse(body);
```
Davlis: I believe we should create some utils/depedency that would make a `http` calls and parse data, because we have highly unrelated code to plugin specific thing.

Maybe we can open-source libary?

src/props/index.js
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      usersSlackIdsToPropsIds = await propsService.getPropsUsers();
+
+    } catch (err) {
+      res.send('It seems like I can\'t get data from props server :( - Try it here: https://netguru.props.devguru.co');
+      return propsService.serverError(err);
+    }
+
+    let unrecognisableUsers = propsService.getUnrecognisableUsers(usersToGivePropsTo);
+    if (!propsService.userIsRecognisable(propserSlackId) || unrecognisableUsers.length > 0) {
```
Davlis: Maybe we could delegate this check to service method with some descriptive name?

What do you think?

src/props/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+      try {
+        await propsService.getUsersFromServer();
```
Davlis: Why do we need that call if we performed `await propsService.getPropsUsers();` earlier?

src/props/index.js
```diff
@@ -0,0 +1,74 @@
...
+        res.send('It seems like I can\'t get data from props server :( - Try it here: https://netguru.props.devguru.co');
+        return propsService.serverError(err);
+      }
+
+      if (!propsService.userIsRecognisable(propserSlackId)) {
+        return res.send('You sure you are signed up for props? Try it here: https://netguru.props.devguru.co');
+      }
+      unrecognisableUsers = propsService.getUnrecognisableUsers(usersToGivePropsTo);
+      if (unrecognisableUsers.length > 0) {
+        const names = unrecognisableUsers
```
Davlis: Please consider making some serializer for that message, or we could rely on some service method or `mustache` template? What do you think about it?

src/props/index.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    let textParser = res.match[0].split(/give_props (@\w+ )+/);
+
+    const userIds = usersToGivePropsTo.map(user => usersSlackIdsToPropsIds[user.id]);
+    const propserId = usersSlackIdsToPropsIds[propserSlackId];
+    const body = textParser[textParser.length - 1];
+
+    try {
+      propsService.giveProp(propserId, userIds, body);
```
Davlis: We should wait for call to finish. (await)

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
+class PropsService {
+  constructor(robot) {
+    this.robot = robot;
+    this.api = robot.brain.propsAPI;
+    // SLACK_UID:PROPS_ID
+    this.propsUsers;
+  }
+  async giveProp(propserId, userIds, body) {
+   return await this.api.giveProp(propserId, userIds, body);
```
Davlis: It's unnecessary to wait for async function when you returning it

Please consider returning
```javascript
 return this.api.giveProp(propserId, userIds, body);
```

src/props/props.service.js
```diff
@@ -0,0 +1,60 @@
...
+    return res.message.mentions
+      .filter(user => user.type === 'user');
+  }
+
+  serverError(err) {
+    this.robot.logger.error(err);
+  }
+
+  parseUsers(data) {
+    if (!(data instanceof Array))
```
Davlis: Could you provide some information what does it do? I don't get it

scripts/props.js
```diff
@@ -0,0 +1,22 @@
+// Description:
+//  If you are authenticated it allows you to send props
+//
+//
+// Commands:
+//  bob give_props <user/s> <text> - give user/s the props with text
```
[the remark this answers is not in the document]
lubieniebieski: we need to check it on CI-level - it should be relatively easy - you can try to implement this @Davlis - I've mentioned this some time ago - if you need help just ping me!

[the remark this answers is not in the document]
Davlis: @lubieniebieski I implemented this on CI-level long time ago, but I didn't handled scenario when someone add documentation to script file, but didn't run docs script 😅My bad. 

Fixed on #24

src/props/api.js
```diff
@@ -0,0 +1,65 @@
+const PROPS_URL = 'https://netguru.props.devguru.co';
+const PROPS_API_PATH = '/api/v1';
+const PROPS_API_TOKEN = process.env.PROPS_API_TOKEN;
```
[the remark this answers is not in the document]
Getriax: Added 
> PROPS_API_TOKEN 
> PROPS_URL 

to .env.example

src/props/api.js
```diff
@@ -0,0 +1,65 @@
...
+
+class API {
+  constructor(robot) {
+    this.robot = robot;
+  }
+
+  getAllUsersFromProps() {
+
+    return new Promise((resolve, reject) => {
+      this.robot.http(PROPS_URL)
```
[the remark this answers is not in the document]
Getriax: Unfortunately, robot.http is based on https://github.com/technoweenie/node-scoped-http-client, which does not support promises. In that case I'll use axios because it will also resolve the problem with parsing the api data to json.

scripts/props.js
```diff
@@ -0,0 +1,22 @@
...
+const props = require('../src/props');
+const API = require('../src/props/api');
+const PropsService = require('../src/props/props.service');
+
+module.exports = robot => {
+
+  robot.brain.propsAPI = new API(robot);
+  robot.brain.propsService = new PropsService(robot);
+
+
```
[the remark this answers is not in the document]
Getriax: Done!

src/props/props.service.js
Getriax: Done!

src/props/index.js
Getriax: `getPropsUsers()` returns the cached (if available) users or makes a request to API for users.

The above if statement 
`!propsService.userIsRecognisable(propserSlackId) || unrecognisableUsers.length > 0`
is checking if there is an id for the user from the props database. 

if not it means that probably the cached version of users is not up to date, so the 
`await propsService.getUsersFromServer();`
is making sure that the data is updated.

src/props/index.js
Getriax: Done!

src/props/api.js
Getriax: Actually, I had this extension - disabled, unfortunately. Already enabled and fixed eslint warnings.

src/props/props.service.js
Getriax: Done!

test/factories/PropsFactory.js
Getriax: Thanks for noticing, fixed!

test/props.spec.js
Getriax: Done!

test/props.spec.js
Getriax: eslint... 😅
Fixed!

test/props.spec.js
Getriax: Done!

src/props/props.service.js
Getriax: I fetch users and then cache them, but if a new user tries to give_props, and there is not a record of him in the cached version, then there will be a fetch from the API to get updated users data. But still, If we want to make a fetch from the server every time I can do that, just let me know 😅

src/props/props.service.js
Getriax: Done!

src/props/api.js
Getriax: Removed!

src/props/api.js
Getriax: Switching to axios will fix that as well.

src/props/index.js
Getriax: const  someoneIsUnrecognisable = (propserSlackId, usersToGivePropsTo,propsService) => {
	   const  unrecognisableUsers = propsService.getUnrecognisableUsers(usersToGivePropsTo);
	   return  !propsService.userIsRecognisable(propserSlackId) || unrecognisableUsers.length  >  0;
	};

**Will do?**

scripts/props.js
Getriax: Done!

src/props/index.js
Getriax: Done!

src/props/index.js
Getriax: Done!

src/props/index.js
```diff
@@ -0,0 +1,21 @@
...
+    res.send(new MessageRenderer('props/help').render());
+  };
+
+  robot.respond(/give_props (.*)/i, (res) => {
+    const { propsService } = robot.brain;
+    const { logger } = robot;
+    const multiplexer = new RespondMultiplexer();
+    const propsController = new PropsController(propsService, logger);
+
+    multiplexer.addResponse(/give_props (@.+)+ ((?!@).+)/i, propsController.giveProps.bind(propsController));
```
krzyurb: Why you `.bind` here?

Maybe try with:
```
multiplexer.addResponse(/give_props (@.+)+ ((?!@).+)/i, propsController.giveProps);
...
action.call(propsController, res, match);
```

src/props/index.js
```diff
@@ -0,0 +1,21 @@
+const { MessageRenderer, RespondMultiplexer } = require('./../../utils');
+const PropsController = require('./props.controller');
+
+module.exports = (robot) => {
+  const propsHelp = (res) => {
```
krzyurb: Consider moving it into `PropsController`

src/props/api.js
```diff
@@ -0,0 +1,50 @@
+const { PROPS_URL, PROPS_API_TOKEN } = process.env;
+const PROPS_API_PATH = '/api/v1';
+const GET_ALL_USERS_API_PATH = '/users';
+const GIVE_PROPS_API_PATH = '/props';
+const axios = require('axios');
```
krzyurb: Consider moving require (`const axios = require('axios');`) over const declarations

src/props/api.js
```diff
@@ -0,0 +1,50 @@
+const { PROPS_URL, PROPS_API_TOKEN } = process.env;
+const PROPS_API_PATH = '/api/v1';
+const GET_ALL_USERS_API_PATH = '/users';
+const GIVE_PROPS_API_PATH = '/props';
```
krzyurb: @Davlis suggested to move `PROPS_URl` and `API_TOKEN` to env. How about moving other things into separated  `config/props.config.json` file?

src/props/props.controller.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if (!this.propsService.userIsRecognisable(propserSlackId)) {
+      return res.send('You sure you are signed up for props? I can\'t find you :(');
+    }
+
+    const unrecognisableUsers = this.propsService.getUnrecognisableUsers(usersToGivePropsTo)
+      .map(user => user.info.real_name);
+
+    if (unrecognisableUsers.length > 0) {
+      const response = new MessageRenderer('props/unrecognisable', { users: unrecognisableUsers }).render();
```
krzyurb: Consider splitting it to avoid having lines longer than 100

```
const renderer = new MessageRenderer('props/unrecognisable', { users: unrecognisableUsers })
return res.send(renderer.render());

src/props/index.js
Getriax: Agreed, `.call` is better here. Fixed!

src/props/index.js
Getriax: Done!

src/props/api.js
Getriax: Done!

src/props/api.js
Getriax: Good idea, done!

src/props/props.controller.js
Getriax: Done!

PR #26, opened 2018-08-22
Upgrade CircleCI plugin

Davlis:
**Pull request information**

- Add workflow support
- Add humanize duration support when asking for statistics
- Add username pass obligation
- Remove depracated commands and it's tests
- Fixes for documentation


templates/circle/help.message
```diff
@@ -1,9 +1,9 @@
...
->`circle retry bob-the-bot master`
->`circle cancel bob-the-bot master`
\ No newline at end of file
+>`circle list netguru/bob-the-bot feature/reminder`
+>`circle stats netguru/bob-the-bot feature/github`
+>`circle last netguru/bob-the-bot feature/profile`
+>`circle clean netguru/bob-the-bot`
+>`circle start netguru/bob-the-bot master`
+>`circle retry netguru/bob-the-bot 274`
+>`circle cancel netguru/bob-the-bot 214`
```
krzyurb: Consider changing values to more abstract things like
```
circle cancel owner/repository branch-id
circle retry owner/repository branch-name
```

It's easer to understand for future users, and it hides `netguru` names from project :)

src/circleci/index.js
```diff
@@ -10,16 +10,11 @@ module.exports = (robot) => {
...
+    multiplexer.addResponse(/circle stats (\S+)\/(\S+)(?:\s*)(\S*)/i, controller.circleStats);
+    multiplexer.addResponse(/circle last (\S+)\/(\S+)(?:\s*)(\S*)/i, controller.circleLast);
 
-    multiplexer.addResponse(/circle clean (.*)/i, controller.circleClean);
-
-    /* Only for CircleCI 1.0 https://discuss.circleci.com/t/triggering-workflows-by-api/22673/5 */
-    multiplexer.addResponse(/circle retry (.*) (.*)/i, controller.circleRetry);
-    multiplexer.addResponse(/circle cancel (.*) (.*)/i, controller.circleCancel);
-    multiplexer.addResponse(/circle start (.*) (.*)/i, controller.circleStart);
+    multiplexer.addResponse(/circle clean (.+)\/(.+)/i, controller.circleClean);
```
krzyurb: Consider using `circle clean (\S+)\/(\S+)`. It will prevent from parsing things like `circle clean aaa bbb/ccc ddd`

test/circleci.spec.js
```diff
@@ -12,78 +12,88 @@ const CircleCiService = require('../src/circleci/circleci.service');
...
-  retryBuild: async () => {},
   clearBuildCache: async () => {},
 };
 
 describe('CircleCi test suite', () => {
-  const buildToString = (build) => {
-    let buildRes = `> #${build.build_num}) :red_circle: 1.00s *${build.workflows.job_name}*`;
-    buildRes += ` [${build.branch}/${build.vcs_revision.substring(0, 7)}]`;
-    buildRes += ` ${build.committer_name}: ${build.subject}\n\n`;
+  const buildToString = (build, project, branch) => {
```
krzyurb: Why you prefer to use `buildToString` function instead of `MessageRenderer`? If anything will change in message schema, it will be necessary to change it here as well.

test/circleci.spec.js
```diff
@@ -136,52 +146,55 @@ describe('CircleCi test suite', () => {
...
     it('should inform about last build #1', async () => {
-      const build1 = BuildFactory({ failed: true, build_time_millis: 1000 });
-      const build2 = BuildFactory({ failed: false, build_time_millis: 1000 });
+      const build1 = BuildFactory({ failed: true, build_time_millis: 1000, build_num: 2 });
+      const build2 = BuildFactory({ failed: false, build_time_millis: 1000, build_num: 1 });
 
-      let build1Res = `> #${build1.build_num}) :red_circle: 1.00s *${build1.workflows.job_name}*`;
-      build1Res += ` [${build1.branch}/${build1.vcs_revision.substring(0, 7)}]`;
-      build1Res += ` ${build1.committer_name}: ${build1.subject}\n\n`;
+      let build1Res = 'Build execution for `bob-the-bot`/`master`\n\n';
```
krzyurb: The same question as with `buildToString ` function.

test/circleci.spec.js
Davlis: @krzyurb If I use `MessageRenderer` tests starts to be very trival in terms of output sent by `hubot`. 

Consider situation when you ask `circle list netguru/bob-the-bot` but we forgot to add message definition 😅 We would only catch that error when requesting to `Bob` not when testing code

test/circleci.spec.js
Davlis: 🆙

src/circleci/index.js
Davlis: Good catch!

test/circleci.spec.js
krzyurb: Ok, you are right 👍

PR #28, opened 2018-08-23
Github hygiene: show pull requests with unfinished reviews

krzyurb:
Command `github hygiene <repository>` shows all pull requests with **started** but unfinished reviews.


package.json
```diff
@@ -14,15 +14,16 @@
   "dependencies": {
     "@octokit/rest": "^15.9.5",
     "@slack/client": "^4.3.1",
+    "asyncro": "^3.0.0",
```
Davlis: Why do we need `asyncro`?

krzyurb: I needed it [here](https://github.com/netguru/bob-the-bot/pull/28/files#diff-0a234c33295f485578ff291595cd91a3R65) to resolve mapping collection using async function.

PR #29, opened 2018-08-24
Improvments Feature/props

Getriax:
**Fixes and changes:**

- Now bob will notify a user if there is no PROPS_API_TOKEN available, (@krzyurb idea 👋)
- Bob will also notify if user tries to props h[im/er]self.
- Test cases for these situations


Davlis:
@Getriax Please resolve conflicts

PR #30, opened 2018-08-24
Improvement of props script

Getriax:
**Fixes and changes:**

- Now bob will notify a user if there is no PROPS_API_TOKEN available, (@krzyurb idea 👋)
- Bob will also notify if user tries to props h[im/er]self.
- Test cases for these situations


src/props/props.controller.js
```diff
@@ -9,14 +9,27 @@ class PropsController {
     this.logger = logger;
   }
 
+  propsApiTokenIsNotAvailable() {
+    const { PROPS_API_TOKEN } = process.env;
+    return PROPS_API_TOKEN === '' || PROPS_API_TOKEN === undefined;
+  }
+
   async giveProps(res) {
+    if (this.propsApiTokenIsNotAvailable()) {
```
krzyurb: Why not simply `if (!process.env.PROPS_API_TOKEN)` ?

src/props/props.controller.js
```diff
@@ -9,14 +9,27 @@ class PropsController {
     this.logger = logger;
   }
 
+  propsApiTokenIsNotAvailable() {
+    const { PROPS_API_TOKEN } = process.env;
+    return PROPS_API_TOKEN === '' || PROPS_API_TOKEN === undefined;
```
Davlis: Consider changing to
```javascript
return !PROPS_API_TOKEN;
```

test/props.spec.js
```diff
@@ -31,13 +31,28 @@ describe('give_props <users> <text>', () => {
       ]);
 
       sinon.stub(propsAPI, 'giveProp').resolves([]);
+
+      process.env.PROPS_API_TOKEN = 'example';
```
Davlis: You can also define this environment in `.env.test.example` and `.env.test`

test/props.spec.js
```diff
@@ -31,13 +31,28 @@ describe('give_props <users> <text>', () => {
...
     });
 
     afterEach(() => {
       sinon.restore();
       room.destroy();
     });
 
+    it('Should notify a user about missing api token', async () => {
+      process.env.PROPS_API_TOKEN = '';
```
Davlis: In my opinion, stubing this is much more cleaner, what do you think?

src/props/props.controller.js
Getriax: I think that a function is more descriptive, that is why I've chosen this way.
Of course, both examples have their pros.

src/props/props.controller.js
Getriax: Done!

test/props.spec.js
Getriax: Fixed!

test/props.spec.js
Getriax: Done!

test/props.spec.js
```diff
@@ -17,6 +17,7 @@ describe('give_props <users> <text>', () => {
     let room = null;
     let propsService;
     let propsAPI;
+    console.log('TOKEN', process.env.PROPS_API_TOKEN);
```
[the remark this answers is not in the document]
Getriax: Left it by mistake, fixed!

src/props/props.controller.js
```diff
@@ -9,14 +9,27 @@ class PropsController {
     this.logger = logger;
   }
 
+  propsApiTokenIsNotAvailable() {
```
Davlis: Consider returning `PROPS_API_TOKEN` directly from `process.env` 

```javascript
propsApiTokenIsNotAvailable() {
  return !process.env.PROPS_API_TOKEN;
}
```

Getriax: Done!

PR #32, opened 2018-08-24
Improve support for documentation comparison

Getriax:
**Changes:**
- Created bin/check scipt that checks that the `yarn run docs` command was used before commit.
- Script will prompt a message informing what to do in case of a failure.


Davlis:
We have talked on Slack about it, but I'll leave it here to not lost this topic.

Please try to implement more human readable error when docs-parser throws one (--strict flag)

PR #35, opened 2018-08-27
RespondMultiplexer error handling

Getriax:
**Description:**
RespondMultiplexer should handle the errors thrown by the controller functions that are stored in the responses object. It will allow to omit try...catch blocks in the controller.

---

**Changes:**
* **RespondMultiplexer** now can handle the errors thrown by controller,
* **props.errors.js** as an example file how it migh look with what message should user get in case of an error,
* **RespondMulitplexer** has now the exec command that chained with choose command exetutes the chosen controller function.


src/props/props.controller.js
```diff
@@ -55,13 +49,8 @@ class PropsController {
...
 
-    try {
-      await this.propsService.giveProp(propserId, userIds, body);
-      return res.send('Success! Props given! 👌');
-    } catch (err) {
-      this.logger.error(err);
-      return res.send('It seems like I can\'t get post data to props server :(');
-    }
+    await this.propsService.giveProp(propserId, userIds, body);
+    return res.send('Success! Props given! 👌');
```
Davlis: You don't need return

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +26,37 @@ class RespondMultiplexer {
...
+      if (functionIsAsync(action)) {
+        await action.call(controller, res, match);
+      } else {
+        action.call(controller, res, match);
+      }
+    } catch (err) {
+      this.loggError(err);
+      const { message } = err;
+      const response = this.messages[message] || this.messages.default;
+      return res.send(response);
```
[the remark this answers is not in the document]
Davlis: I think that this should be separated to another function

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +26,37 @@ class RespondMultiplexer {
     };
   }
 
+  loggError(err) {
+    if (process.env.NODE_ENV !== 'test') {
```
Davlis: Consider to remove this if statement.
You can always set proper logger level in `.env.test`

https://www.npmjs.com/package/log - **Log Levels section**

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +26,37 @@ class RespondMultiplexer {
...
     const response = this.responses.find(c => c.trigger.exec(text)) || this.default;
     const match = response.trigger.exec(text);
-    return { action: response.action, match };
+    this.chosen = { action: response.action, match };
+    return this;
+  }
+
+  async exec(controller, res) {
+    if(!this.chosen.action) {
+      console.error('Cannot exectute nothing chosen');
```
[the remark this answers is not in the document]
Davlis: Why not logger?

src/props/props.errors.js
```diff
@@ -0,0 +1,16 @@
+
+const getMessage = error => error.message;
```
Davlis: This is wrong approach for getting type of error.

Please consider creating your own derieved class of Error, or use some pre-defined error-responses libary, for example https://www.npmjs.com/package/boom

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -1,10 +1,18 @@
+
+const functionIsAsync = (fn) => {
```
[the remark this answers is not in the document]
Davlis: You no need this function

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +26,37 @@ class RespondMultiplexer {
...
+  }
+
+  async exec(controller, res) {
+    if(!this.chosen.action) {
+      console.error('Cannot exectute nothing chosen');
+      return;
+    }
+    const { action, match } = this.chosen;
+    try {
+      if (functionIsAsync(action)) {
```
Davlis: You may `await` for ordinary functions too

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +26,37 @@ class RespondMultiplexer {
...
+        await action.call(controller, res, match);
+      } else {
+        action.call(controller, res, match);
```
Davlis: What do you think about introducing `auto-bind` and use it in controllers? We could avoid using `call` and passing `controller` to `RespondMultiplexer`

https://www.npmjs.com/package/auto-bind

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +26,37 @@ class RespondMultiplexer {
...
   choose(text) {
     const response = this.responses.find(c => c.trigger.exec(text)) || this.default;
     const match = response.trigger.exec(text);
-    return { action: response.action, match };
+    this.chosen = { action: response.action, match };
+    return this;
+  }
+
+  async exec(controller, res) {
+    if(!this.chosen.action) {
```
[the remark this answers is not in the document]
Davlis: I strongly believe that this if statement is not necessary.

Maybe we could modify code to be sure that `this.chosen.action` is always defined?

utils/RespondMultiplexer/src/RespondMultiplexer.js
Getriax: That is a great idea, implemented!

src/props/props.errors.js
Getriax: Corrected!

src/props/props.errors.js
```diff
@@ -0,0 +1,16 @@
...
+
+const error = {
```
krzyurb: It's possible that in future there will be more errors. Consider changing name to `errors`.

src/props/props.errors.js
```diff
@@ -0,0 +1,16 @@
...
+  SERVER: new Error('Connection with server error'),
+};
+
+const messages = {
```
krzyurb: I'm not sure is it the best idea to store messages like this. Maybe it could be an error object propery?

We could have common Bot error definition inside `utils` , so we will be able to use it in other controllers
```
class BotError extends Error {
  constructor(errorMessage, botMessage) {
    super(errorMessage);
    this.botMessage = botMessage;
  }
}
```

And inside `props.errors.js` you can do something like 
```
const = errors {
  server: new BotError('Connection with server error', 'Ups... something went wrong try again later!'),
  ...
}
````

Than if you will catch an error you can simply get message for bot response using:
```
res.send(error.botMessage);
```

What do you thing about it?

File: utils/RespondMultiplexer/src/RespondMultiplexer.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (Getriax) [reply utils/RespondMultiplexer/src/RespondMultiplexer.js]: That is correct, the default action is always available 😅

File: src/props/props.controller.js (hunk 1 above)
Comment (Getriax) [reply src/props/props.controller.js]: Corrected

File: utils/RespondMultiplexer/src/RespondMultiplexer.js (hunk 3 above)
Comment (Getriax) [reply utils/RespondMultiplexer/src/RespondMultiplexer.js]: Removed

File: src/props/props.errors.js (hunk 11 above)
Comment (Getriax) [reply src/props/props.errors.js]: Yes, I've implemented this solution already 👍

**Comment by Getriax** [comment]
@Davlis @krzyurb Fixes done.

File: test/props.spec.js [hunk 12]
```diff
@@ -306,19 +307,19 @@ describe('give_props <users> <text>', () => {
...
       }
     });
-    it('Should throw an error if cannot get data', async () => {
-      sinon.stub(axios, 'post').returns(Promise.reject(new Error('error')));
+    it('Should throw an error if cannot post data', async () => {
+      sinon.stub(axios, 'post').rejects();
 
-      const expected = new Error('POST DATA ERROR! Error: error');
+      const expected = errors.SERVER;
       try {
```
Davlis: Test is okay, but I believe that using mocha tools to test if proper error was thrown is easier to understand and more descriptive

https://stackoverflow.com/questions/14966821/testing-for-errors-thrown-in-mocha

src/props/index.js
```diff
@@ -1,17 +1,21 @@
...
 const PropsController = require('./props.controller');
 
 module.exports = (robot) => {
-  robot.respond(/give_props (.*)/i, (res) => {
+  robot.respond(/give_props (.*)/i, async (res) => {
     const { propsService } = robot.brain;
     const { logger } = robot;
-    const multiplexer = new RespondMultiplexer();
-    const propsController = new PropsController(propsService, logger);
+    const multiplexer = new RespondMultiplexer(logger, 'Something is wrong with props, try again later!');
```
Davlis: I don't think that passing string argument near service is good idea.

Please consider creating some methond in RespondMultiplexer where you could pass error messasges / set default

src/props/index.js
```diff
@@ -1,17 +1,21 @@
...
+    const propsController = new PropsController(propsService);
+    autoBind(propsController);
```
Davlis: Consider using autoBind in `props.controller.js`

src/props/props.controller.js
```diff
@@ -1,67 +1,17 @@
...
-    try {
-      await this.propsService.giveProp(propserId, userIds, body);
-      return res.send('Success! Props given! 👌');
-    } catch (err) {
-      this.logger.error(err);
-      return res.send('It seems like I can\'t get post data to props server :(');
-    }
+    const { userIds, propserId, body } = await this.propsService.getPropsData(res, this.logger);
+    await this.propsService.giveProp(propserId, userIds, body);
+    return res.send('Success! Props given! 👌');
```
Davlis: Consider adding more new lines to separate different logic

Davlis: Consider removing return

src/props/props.service.js
```diff
@@ -25,6 +26,49 @@ class PropsService {
     return this.propsUsers;
   }
 
+  async getPropsData(res) {
+    PropsService.checkIfTokenIsAvailable();
```
Davlis: This function name (checkIfTokenIsAvailable) suggests that it would return true/false but throws an error

I would highly recommend to change name of this function OR change some logic here (wrap this in try-catch...etc)

src/props/props.service.js
```diff
@@ -25,6 +26,49 @@ class PropsService {
...
+
+    const propserSlackId = PropsService.getUserSlackId(res);
+    if (!propserSlackId) {
+      throw errors.AUTHENTICATION;
+    }
+
+    const usersToGivePropsTo = PropsService.getUsersFromMentions(res);
+    PropsService.checkIfPropserIsTryingToPropSelf(propserSlackId, usersToGivePropsTo);
```
Davlis: Same situation as above

You can also try to implement  similar [function](https://github.com/Davlis/blogger-api/blob/7b73e7cbbd0363a19c03de33610b0d79171ee749/src/utils.js#L11) to our scenario

Example:

```javascript
assertOrThrow(PropsService.checkIfPropserIsTryingToPropSelf(propserSlackId, usersToGivePropsTo), Error)
```

src/props/props.service.js
```diff
@@ -25,6 +26,49 @@ class PropsService {
...
+
+    const usersSlackIdsToPropsIds = await this.getUsersFromServer();
+
+    this.checkIfSomeoneIsUnrecognisable(propserSlackId, usersToGivePropsTo);
+
+    const textParser = res.match[0].split(/give_props (@\w+ )+/);
+    return {
+      userIds: usersToGivePropsTo.map(user => usersSlackIdsToPropsIds[user.id]),
+      propserId: usersSlackIdsToPropsIds[propserSlackId],
+      body: textParser[textParser.length - 1],
```
Davlis: I know that this is not a case for RespondMultiplexer, but I found this now

Please consider separate extracting props body to some external function with some descriptive name

src/props/props.service.js
```diff
@@ -25,6 +26,49 @@ class PropsService {
...
+      throw errors.PROPSER_404;
+    }
+
+    const unrecognisableUsers = this.getUnrecognisableUsers(usersToGivePropsTo)
+      .map(user => user.info.real_name);
+
+    if (unrecognisableUsers.length > 0) {
+      throw errors.USERS_404(unrecognisableUsers);
+    }
+    return false;
```
Davlis: You mix throwing error logic with returning values

src/props/props.errors.js
```diff
@@ -0,0 +1,17 @@
+const { BotError, MessageRenderer } = require('../../utils');
+
+const errors = {
```
Davlis: I like this structure 👍

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +21,26 @@ class RespondMultiplexer {
...
 
+  loggError(err) {
+    this.logger.debug(err);
+  }
+
   choose(text) {
     const response = this.responses.find(c => c.trigger.exec(text)) || this.default;
     const match = response.trigger.exec(text);
-    return { action: response.action, match };
+    this.chosen = { action: response.action, match };
```
Davlis: Please consider adding more new lines

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -1,10 +1,13 @@
 class RespondMultiplexer {
-  constructor() {
+  constructor(logger, defaultUserErrorMessage = 'Something went wrong :(') {
```
Davlis: I think that you should hardcode this defaultErrorMessage to class body, not to constructor

test/props.spec.js
Getriax: `chai-as-promised` 😅

src/props/index.js
Getriax: Fixed!

src/props/index.js
Getriax: Moved!

src/props/props.controller.js
Getriax: Fixed!

src/props/props.service.js
Getriax: Went with `assertOrThrow(PropsService.checkIfPropserIsTryingToPropSelf(propserSlackId, usersToGivePropsTo), Error)` suggestion

src/props/props.service.js
Getriax: Fixed!

utils/RespondMultiplexer/src/RespondMultiplexer.js
Getriax: Added!

utils/RespondMultiplexer/src/RespondMultiplexer.js
Getriax: Moved to the class body!

Getriax:
@Davlis Fixes done! 🔨

src/props/props.service.js
```diff
@@ -1,3 +1,10 @@
+const { errors } = require('./props.errors');
+
+const assertOrThrow = (check, error) => {
```
Davlis: Please consider moving this to utils, and create additional unit tests for that utility

src/props/props.service.js
```diff
@@ -25,6 +28,47 @@ class PropsService {
     return this.propsUsers;
   }
 
+  async getPropsData(res) {
+    await assertOrThrow(PropsService.isTokenAvailable, errors.API_TOKEN);
+
+    const propserSlackId = PropsService.getUserSlackId(res);
+    if (!propserSlackId) {
```
Davlis: You can use `assertOrThrow`

Getriax:
@Davlis Fixed

test/utils/assertOrThrow.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+const { expect } = chai;
+const { assertOrThrow } = require('../../utils');
+
+const dummy = value => value;
+
+const asyncDummy = async (value) => {
+  await process.nextTick(() => {});
+  return value;
+};
```
Davlis: You can write it as
```javascript
const asyncDummy = async value => value;
```

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +21,33 @@ class RespondMultiplexer {
     };
   }
 
+  setDefaultUserErrorMessage(text) {
+    this.defaultUserErrorMessage = text;
+  }
+
+  loggError(err) {
```
Davlis: Shouldn't it be `logError`?

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +21,33 @@ class RespondMultiplexer {
...
+
+    this.chosen = { action: response.action, match };
+
+    return this;
+  }
+
+  async exec(res) {
+    const { action, match } = this.chosen;
+    try {
+      await action(res, match);
```
Davlis: You may return result of `await action(res, match)` and  therefore you won't need changing tests for `RespondMultiplexer`

utils/RespondMultiplexer/src/RespondMultiplexer.js
```diff
@@ -18,10 +21,33 @@ class RespondMultiplexer {
...
+    } catch (err) {
+      this.loggError(err);
+
+      const response = err.userMessage || this.defaultUserErrorMessage;
```
Davlis: I'm not sure, but maybe it's better to check whether error is instance of BotError ? 
I'm just little confused when I see some field that I don't know what is it's purpose :D

utils/RespondMultiplexer/src/RespondMultiplexer.js
Getriax: Fixed!

test/utils/assertOrThrow.spec.js
Getriax: Done!

utils/RespondMultiplexer/src/RespondMultiplexer.js
Getriax: Fixed!

PR #36, opened 2018-08-29
Implementation of interactive messages and questions plugin

krzyurb:
### Task description

Implementation of questions plugin. It allows one user to ask others a question through bot.

### Simplified flow

1. User sends to Bob message `ask @other.user for what's up?`
2. Bob sends the question as a private message to the mentioned user(s).
3. If users will answer a question, bot will send them back to the author of question as private message.

#### Sending a question

gif01

#### Answering a question

gif02

### Interactive messages

This plugin was made using slack _interactive components_. They allow us to create interactive messages with buttons, dialog windows, etc. 

### How to make interactive components work?

At first it's necessary to activate `Interactivity` for bot app in slack apps manager and to provide _url_ of deployed bot.

screen shot 2018-08-29 at 14 29 31

While developing on local machine i recommend using `ngrok`.

### Simplified interactive messages workflow

1. To display message with buttons we need to prepare message in _JSON_ format
2. After pressing a button, slack UI is making a `POST` request to defined earlier request url.
3. Hubot contains _express_ server which is consuming slack request. Each action has own, unique `callback_id` which allows us to identify action to handle.
4. After handling action - we have to send response with proper status.

More: [api.slack.com/interactive-messages](https://api.slack.com/interactive-messages)

### JIRA

[TECH-94](https://netguru.atlassian.net/browse/TECH-94)


Davlis:
@krzyurb  Please include some description

src/slackActions/index.js
```diff
@@ -0,0 +1,20 @@
...
+const { askQuestion, submitAnswer } = require('./question');
+
+module.exports = (robot) => {
+  robot.router.post(process.env.HUBOT_SLACK_ACTIONS_ENDPOINT, (req, res) => {
+    try {
+      const actionJSONPayload = JSON.parse(req.body.payload);
+
+      const multiplexer = new CallbackMultiplexer();
+      multiplexer.addResponse('ask-question', askQuestion);
+      multiplexer.addResponse('answer-question', submitAnswer);
```
Getriax: If you would consider using RespondMultiplexer here with the regex instead of strings it should also work. 
Here is why:

`multiplexer.addResponse('ask-question', askQuestion);`
is triggered by choosing `actionJSONPayload.callback_id` which could also be `ask-question`

In CallbackMultiplexer you are comapring it with `c.callbackId === text`.

In case of using RespondMultiplexer it would look like:
`multiplexer.addResponse(/ask-question/, askQuestion);`
Again choose provided with the same value would be triggered.
`c.trigger.exec(text)`

So the same action would be retrieved in both cases.

package.json
```diff
@@ -14,19 +14,20 @@
...
     "hubot": "3",
     "hubot-help": "^0.2.2",
     "hubot-heroku-keepalive": "^1.0.3",
-    "hubot-redis-brain": "^0.0.4",
     "hubot-slack": "^4.5.3",
     "humanize-duration": "^3.15.1",
     "lodash": "^4.17.10",
     "moment": "^2.22.2",
     "mustache": "^2.3.1",
+    "querystring": "^0.2.0",
```
Getriax: Overall `querystring` is a default node module, so there is no need to add it to the dependencies.

src/slackActions/index.js
krzyurb: You are right. `RespondMultiplexer` will be fine here.

krzyurb:
@Getriax done!

Davlis:
@krzyurb when `HUBOT_SLACK_ACTIONS_ENDPOINT` env variable is not specifed then not every test pass

src/slackActions/question.js
```diff
@@ -0,0 +1,73 @@
+const axios = require('axios');
+const qs = require('querystring');
+
+const askQuestion = async (robot, payload, res) => {
```
Davlis: `robot` is never used

README.md
```diff
@@ -249,3 +249,9 @@ If you are authenticated it allows you to send props
 |Trigger|Description|
 |---|---|
 |bob give_props \<user/s> \<text>|give user/s the props with text|
+## question.js
+Script to ask people for something and collect answers
+
+## slackActions.js
```
Davlis: You can mention about HUBOT_SLACK_ACTIONS_ENDPOINT environment variable

README.md
```diff
@@ -249,3 +254,9 @@ If you are authenticated it allows you to send props
...
+Slack http actions handler useed by slack interactive messages
```
Davlis: Typo :)

krzyurb: Good eye!

krzyurb:
@Davlis Typo fix done

src/question/index.js
```diff
@@ -0,0 +1,15 @@
+module.exports = (robot) => {
+  robot.respond(/ask (.*) for (.*)/i, async (res) => {
```
Davlis: You can try to use such RegExp 
```javascript
/ask\s+((@\S+\s+)+)\s+for\s(.+)/i
```

It would match `ask    @someone @someone     for something`

But no `ask someone for something`

krzyurb: Nice idea. But your regexp is not matching `ask @someone @someone for something `.
`ask\s+((@\S+\s+)+)for\s(.+)` seems to work better.

Davlis: Nice catch 👍

test/slackActions/question.spec.js
```diff
@@ -0,0 +1,111 @@
+const Helper = require('hubot-test-helper');
+const sinon = require('sinon');
+const axios = require('axios');
+const { expect } = require('chai');
+const request = require('supertest');
+
+const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+const helper = new Helper('../src/slackActions/index.js');
+const slackEndpoint = process.env.HUBOT_SLACK_ACTIONS_ENDPOINT;
```
Davlis: @krzyurb Please stub this 🙏

krzyurb: Unfortunately stubbing it not enough. I have to make:
```
const slackEndpoint = '/slack/interactivity';
process.env.HUBOT_SLACK_ACTIONS_ENDPOINT = slackEndpoint;
```

to be sure that bot will use this variable when using `supertest` to create instance of express server.

I splitted it into two lines to enable using `slackEndpoint` in tests cases.

krzyurb: I moved that variable into `config/slackActions.config.js`

PR #37, opened 2018-09-03
Message renderer improvements

krzyurb:
Implementation of `MessageBuilder` class that allows us to build fancy json custom slack messages using e.g. attachments feature.

**Usage example**

```
const builder = new MessageBuilder(text: "List of your pull requests");
builder.addAttachment({
  author_name: pullRequest.user.login,
  ...
});

...
builder.disableLinksPreview().buildMessage();
```

**Why not [slack-message-builder](https://www.npmjs.com/package/slack-message-builder)?**
I think that in both cases it's necessary to keep eye on slack api when creating new messages. Having own class allows us to add functions like `disableLinksPreview` witch adds more than one message parametr and add attachments by iterating items.

**Tests**

Tests will run properly after [those](https://github.com/Davlis/hubot-test-helper/pull/2) changes! (edit: done!)

**How it looks like**

screen shot 2018-09-03 at 09 39 35


utils/MessageBuilder/src/MessageBuilder.js
```diff
@@ -0,0 +1,35 @@
+const _ = require('lodash');
+
+class MessageBuilder {
+  constructor(params = {}) {
+    this.attachments = [];
+    this.options = params || {};
```
Getriax: The `|| {}` can be omitted here because `params` has the default value of `{}`

PR #39, opened 2018-09-04
Implement improved RespondMultiplexer in github plugin

krzyurb:
Implementation of ResponsMultiplexer changes in github plugin


src/github/github.controller.js
```diff
@@ -38,41 +41,28 @@ class GithubController {
...
-
-        const mappedField = this.ProfileService.mapField('github');
-        const fieldDefined = await this.ProfileService.isProfileFieldDefined(mappedField);
-
-        if (!fieldDefined) {
-          return res.send('_Github_ field doesn\'t exists in your workspace');
-        }
+    if (user) {
+      const userMention = this.ProfileService.extractUserMention(res);
+      await assertOrThrow(userMention, errors.MENTION_PARSE_ERROR, res);
```
Getriax: Here you are passing the `userMention` as the value, so the last parameter `res` can be omitted.
For further notice, there are two options for using assertOrThrow:
* `assertOrThrow(value, error)`
* `assertOrThrow(function, error, functionParams)`

PR #47, opened 2018-09-07
Question plugin: enables asking channel users

krzyurb:
Improvement to the question plugin. It enables to send the question to all channel members.

**Usage example**
```
hubot ask @tom #friends for What is your favorite color?
```

**How it works**

`prepareMentions` method finds all channels mentions, and map them as chanel members mentions. It is necessary to fetch chanel members using `slack api`.


src/question/question.service.js
```diff
@@ -1,10 +1,32 @@
...
 class QuestionService {
-  prepareMentions(mentions, questionBody) {
-    return _.reject(mentions, mention => questionBody.match(new RegExp(mention.info.name)));
+  constructor(client) {
+    this.client = client;
+  }
+
+  async getChannelUsers(channelId) {
+    const result = await this.client.channels.info({ channel: channelId });
+    return _.map(result.channel.members, o => ({ id: o }));
```
Getriax: I am not sure what `o` stands for, could this variable name be more descriptive? If it is `id` you can simplify it to:

    return _.map(result.channel.members, id => ({ id }));

src/question/question.service.js
```diff
@@ -1,10 +1,32 @@
...
+  }
+
+  async prepareMentions(message, questionBody) {
+    const mentions = await map(message.mentions, async (mention) => {
+      if (mention.type === 'conversation') {
+        return this.getChannelUsers(mention.id);
```
Getriax: I think that you are missing `await` in this line.

src/question/question.service.js
```diff
@@ -1,10 +1,32 @@
...
 
 class QuestionService {
-  prepareMentions(mentions, questionBody) {
-    return _.reject(mentions, mention => questionBody.match(new RegExp(mention.info.name)));
+  constructor(client) {
+    this.client = client;
+  }
+
+  async getChannelUsers(channelId) {
+    const result = await this.client.channels.info({ channel: channelId });
```
Getriax: An extra line here would be nice 😄

src/question/question.service.js
krzyurb: According to eslint - `Redundant use of await on a return value.` :)

Getriax: Yes because eslint doesn't allow returning await, you should consider this:

    const channelUsers = await this.getChannelUsers(mention.id);
    return channelUsers;

krzyurb:
@Getriax Fixes done. I added also [one extra line](https://github.com/netguru/bob-the-bot/pull/47/files#diff-63f952b640271d8088b944fbc3876d34R26) to avoid duplicates.

src/question/question.service.js
krzyurb: >Since the return value of an async function is always wrapped in Promise.resolve , return await doesn't actually do anything except add extra time before the overarching Promise resolves or rejects.

🤔

src/question/question.service.js
```diff
@@ -1,10 +1,35 @@
...
+        return this.getChannelUsers(mention.id);
+      }
+
+      return mention;
+    });
+
+    mentions = _.uniqBy(_.flatten(mentions), 'id');
+
+    return _.reject(mentions, (mention) => {
+      if (_.get(mention, 'info.name')) {
```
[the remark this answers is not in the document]
Davlis: And what logic is behind this part of code 28-31?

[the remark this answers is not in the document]
krzyurb: `mention.info.name`  doesn't work with channels, only with user mentions.
That reject is to remove mentions from question content.  If you will type:
`ask @david for why @krzysztof is so stupid?` it wont send question to `@krzysztof` :)

src/question/question.service.js
```diff
@@ -1,10 +1,35 @@
...
+  }
+
+  async getChannelUsers(channelId) {
+    const result = await this.client.channels.info({ channel: channelId });
+
+    return _.map(result.channel.members, member => ({ id: member }));
+  }
+
+  async prepareMentions(message, questionBody) {
+    let mentions = await map(message.mentions, async (mention) => {
```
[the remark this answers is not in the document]
krzyurb: Yes. It's fetching data from slack api

src/question/question.service.js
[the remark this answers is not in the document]
Davlis: Why don't you split this part to another function :D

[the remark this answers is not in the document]
krzyurb: You name it  🎉

krzyurb:
@Davlis fixes done

PR #58, opened 2018-09-14
Improve confluence plugin response

krzyurb:
Add author, space name and creation date to each confluence result


src/confluence/confluence.service.js
```diff
@@ -21,10 +22,27 @@ class ConfluenceService {
...
       const url = `${this.baseUrl}/wiki${_links.webui}`;
+      const authorUrl = `${this.baseUrl}/people/${history.createdBy.accountId}`;
+      const authorIconUrl = `${this.baseUrl}${history.createdBy.profilePicture.path}`;
 
-      return { title, title_link: url, color: randomColor() };
+      return {
+        title,
+        title_link: url,
+        footer: space.name,
+        author_name: history.createdBy.displayName,
```
Getriax: So I have noticed that the user can be undefined. In that case, I am getting 
`Unknown User (fa2a....534sfd2)`

Moreover looking at the response JSON I saw that `createdBy.displayName` could be equal to 
`Unknown User (fa2a....534sfd2)` while the `createdBy.username` would point to a user.

What do you think about adjusting that a little? 😄

krzyurb: Goooood point!

krzyurb:
@Getriax fixed

src/confluence/confluence.service.js
```diff
@@ -20,28 +20,33 @@ class ConfluenceService {
...
+  authorData(author) {
+    if (author.accountId) {
+      return {
+        author_name: author.displayName,
+        author_link: `${this.baseUrl}/people/${author.accountId}`,
+        author_icon: `${this.baseUrl}${author.profilePicture.path}`,
+      };
+    }
+
+    return {};
```
Getriax: Could you test that case?

PR #60, opened 2018-09-17
[Improvement] Confluence & Props command update

Getriax:
## Task description

Change the command to `conflu|confluence <title>`. Also removed `conflu help` as it will be added with the new **help plugin**.


test/props.spec.js
```diff
@@ -17,7 +17,7 @@ const helper = new Helper('../src/props/index.js');
 const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
```
krzyurb: Please remove one line

PR #64, opened 2018-09-19
Make Bob the Bot advertise devops roles

krzyurb:
**Task description**

Make Bob the Bot advertise DevOps roles. The plugin adds the command to ask for DevOps ninja. It also notifies `#devops` channel each day about all current devops roles.

**Usage example**

screen shot 2018-09-20 at 12 15 42

screen shot 2018-09-20 at 12 16 31

**BotCron**

This PR introduces also CRON actions for the bot. You can define it in scripts `index` according to code below.

`botCron('example-action-label', robot);` - defines CRON action.  Hubot creates event called `example-action-label`, whitch is later handled by `robot.on('example-action-label', ...`

```
const { botCron, dependenciesLocator } = require('./../../utils');

module.exports = (robot) => {
  botCron('example-action-label', robot);

  robot.respond(/example/i, async (res) => {
    // do something
  });

  robot.on('example-action-label', async (params) => {
    const first = params.first; // "blahbla"
    // do something
  });
};
```

Details of an action should be stored in `config/cronActions.config.json`, e.g:

```
  "example-action-label": {
    "description": "blahblabalbhalbh",
    "cron": "0 9 * * 1-5"
    "params" : {
      "first": "blahbla",
      "second": "qweqweqe"
    }
  }
```

You can define params that will be passed to an event handler.

**Jira ticket**
TECH-139


src/devops/devops.dao.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  static decodeContent(content, encoding = 'base64') {
+    const buffer = Buffer.from(content, encoding);
+    return buffer.toString('ascii');
+  }
+
+  async getFileContent(fileName) {
+    const { data } = await this.client.repos.getContent({
+      owner: process.env.DEVOPS_REPO_OWNER, repo: process.env.DEVOPS_REPO_NAME, path: fileName,
```
Getriax: Would you mind splitting that line?

utils/BotCron/src/botCron.js
```diff
@@ -0,0 +1,12 @@
+const { CronJob } = require('cron');
+const { tasks } = require('../../../config/cronTasks.config.json');
+
+module.exports = (task, robot, paramss, autoStart = true) => {
```
Davlis: Just quick note: Cron utility is highly coupled with hubot framework, but for now it's okay. IMO we should try to refactor this in future

config/cronTasks.config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "tasks": {
+    "devops-roles-notify": {
+      "description": "Send message to #devops channel about current roles",
+      "cron": "0 9 * * 1-5"
+    }
+  }
+}
```
Davlis: I don't think that this is bad approach :D but this file forces developers to put every cron task in this file, I think that in future this file may be very large and complex and thus hard to read, maybe we could separate it to scripts config file? 

I know that similar approach is in SlackActionsMultiplexer, but SlackActionsMultiplexer config file is going to be rather small, yeah? :P

scripts/devops.js
```diff
@@ -0,0 +1,19 @@
+// Description
+//  Plugin to make bot advertise devops roles
+//
+// Commands:
+//   hubot devops ninja - shows current devops ninja
+//
+// Notes:
+//   Data about roles is fetched from github repository
```
Davlis: Documentation lacks of `Configuration` field with environment variables

scripts/devops.js
```diff
@@ -0,0 +1,19 @@
+// Description
+//  Plugin to make bot advertise devops roles
```
Davlis: Could you provide more information about this script?

README.md
```diff
@@ -253,6 +253,12 @@ Script to fetch content from organization's confluence
 |Trigger|Description|
 |---|---|
 |conflu \<title>|Displays the available confluence pages/posts|
+## devops.js
+
+
```
Davlis: Something is broken, because lack of description

PR #73, opened 2018-10-11
Redis Client for Bot

krzyurb:
RedisClientFactory method to establish the connection with a database.

Uses [ioredis](https://github.com/luin/ioredis) lib as provider.


krzyurb:
@Getriax You are right - I removed it.

PR #75, opened 2018-10-12
[BOOK-8] Add Users Ids Mapper Service

kozikkam:


krzyurb:
@kozikkam You might have a problem with local git account - it's not linking commits with `kozikkam`. And please open a pull request to `project/book` instead of `master`

PR #76, opened 2018-10-15
Book 8 users id mapper

kozikkam:
Service that will map user emails to Slack IDs. Will be used in service from Task #3 when saving data.


test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+const UsersIdMapperService = require('../../src/slack/UsersIdMapper.service');
+const SlackClientFactory = require('../../src/clients/SlackClientFactory');
+const client = SlackClientFactory();
+
+const errors = require('../../src/slack/slack.errors');
+
+const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+describe('Slack emails to ids mapper', () => {
+  let userEmails = ['super@malpka.pl'];
```
krzyurb: Is it necessary to use an array if you have only one value inside?

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+
+  afterEach(() => sinon.restore());
+
+  describe('requestUsers', () => {
+    it('Should return a valid list of users', async () => {
+      sinon.stub(client.users, 'list').returns(responseOk);
+      const result = await usersIdMapper.requestUsers();
+      expect(result).to.be.an('object');
+      expect(result.members).to.be.an('Array');
+      expect(result.members.length).to.be.above(0);
```
krzyurb: It might be better to expect exact value instead of `.above(0)`. It will be more errorproof.

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      sinon.stub(client.users, 'list').returns(responseOk);
+      const result = await usersIdMapper.requestUsers('ADAWDvvv334WD');
+      expect(result).to.be.an('object');
+      expect(result.members).to.be.an('Array');
+      expect(result.members.length).to.be.above(0);
+    });
+
+    it('Should throw error if no response from server', async () => {
+      sinon.stub(client.users, 'list').returns(undefined);
+      const tested = async () => { await usersIdMapper.requestUsers() };
```
krzyurb: Consider changing it to:

```
const result = await usersIdMapper.requestUsers();
await expect(result).to.be.rejectedWith(errors.SERVER);
```

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+  });
+
+  describe('getMappedSlackUsers', () => {
+    it('Should return users with their ids', async () => {
+      sinon.stub(usersIdMapper, 'requestUsers').returns(responseOkNoMetadata);
+      const result = await usersIdMapper.getMappedSlackUsers();
+      expect(result).to.be.an('Array');
+      expect(result.length).to.be.above(0);
+      expect(result[0]).to.deep.equal({[responseOkNoMetadata.members[0].profile.email]: responseOkNoMetadata.members[0].id});
```
krzyurb: Please split that line to not exceed 100 characters and add a space after { and before }

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+              return resOk;
+            }
+            return responseOkNoMetadata;
+          },
+        },
+      });
+      const result = await usersIdMapper.getMappedSlackUsers();
+      expect(result).to.be.an('Array');
+      expect(result.length).to.be.above(0);
+      expect(result[0]).to.deep.equal({[responseOkNoMetadata.members[0].profile.email]: responseOkNoMetadata.members[0].id});
```
krzyurb: Please split that line and add a space after `{` and before `}`

src/slack/UsersIdMapper.service.js
```diff
@@ -0,0 +1,49 @@
...
+    const mappedUsers = [];
+    let response = await this.requestUsers();
+    for (let i = 0; i < 1; i += 1) {
+      response.members.forEach((member) => {
+        if (this.usersEmails.indexOf(member.profile.email) > -1) {
+          mappedUsers.push({ [member.profile.email]: member.id });
+        }
+      });
+      if (response.response_metadata) {
+        /* eslint-disable no-await-in-loop */
```
krzyurb: Move eslint comments to the top of the file

src/slack/UsersIdMapper.service.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  async requestUsers(cursor = undefined) {
+    const params = {
+      token: process.env.HUBOT_OAUTH_SLACK_TOKEN,
+    };
+    if (cursor) {
+      params.cursor = cursor;
+    }
+    const response = await this.client.users.list(params);
```
krzyurb: Concider removing `params` const and writing it this way
```
const response = await this.client.users.list({
  token: process.env.HUBOT_OAUTH_SLACK_TOKEN,
}
```

src/slack/UsersIdMapper.service.js
```diff
@@ -0,0 +1,49 @@
...
+    this.client = client;
+  }
+
+  async requestUsers(cursor = undefined) {
+    const params = {
+      token: process.env.HUBOT_OAUTH_SLACK_TOKEN,
+    };
+    if (cursor) {
+      params.cursor = cursor;
+    }
```
krzyurb: Please, add an empty line after each `if`. It will make code much cleaner

src/slack/UsersIdMapper.service.js
```diff
@@ -0,0 +1,49 @@
...
+    if (Object.keys(response).length === 0) {
+      throw errors.NO_RESULTS;
+    }
+    return response;
+  }
+
+  async getMappedSlackUsers() {
+    const mappedUsers = [];
+    let response = await this.requestUsers();
```
krzyurb: Add an empty line

src/slack/UsersIdMapper.service.js
```diff
@@ -0,0 +1,49 @@
...
+          mappedUsers.push({ [member.profile.email]: member.id });
+        }
+      });
+      if (response.response_metadata) {
+        /* eslint-disable no-await-in-loop */
+        response = await this.requestUsers(response.response_metadata.next_cursor);
+        /* eslint-enable no-await-in-loop */
+        i -= 1;
+      }
+    }
```
krzyurb: Add an empty line before return

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+const errors = require('../../src/slack/slack.errors');
+
+const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+describe('Slack emails to ids mapper', () => {
+  let userEmails = ['super@malpka.pl'];
+  let usersIdMapper = new UsersIdMapperService(userEmails, client);
+
+  const responseOk = {
+    "ok": true,
```
krzyurb: You don't need quotes here - please remove them according to `[eslint] Unnecessarily quoted property 'ok' found. (quote-props)`

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    "members": [
+        {
+          "id": "P23OA23D9",
```
krzyurb: You have wrong intents in here

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+          "id": "W012A3CDE",
+          "profile": {
+              "email": "hoho@email.example.com"
+          }
+        }
+    ]
+  }
+
+  const responseNotOk = {
+    "ok": false,
```
krzyurb: You don't need quotes here. Are you using `eslint` in yours ide?

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    "error": "invalid_cursor"
+  };
+
+  afterEach(() => sinon.restore());
+
+  describe('requestUsers', () => {
+    it('Should return a valid list of users', async () => {
+      sinon.stub(client.users, 'list').returns(responseOk);
+      const result = await usersIdMapper.requestUsers();
```
krzyurb: Try to split those blocks of code with empty lines. It's good practice to split them into [arrange, action, assert](http://defragdev.com/blog/?p=783) blocks

test/slack/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      expect(result).to.be.an('object');
+      expect(result.members).to.be.an('Array');
+      expect(result.members.length).to.be.above(0);
+    });
+
+    it('Should return a valid list of users from cursor', async () => {
+      sinon.stub(client.users, 'list').returns(responseOk);
+      const result = await usersIdMapper.requestUsers('ADAWDvvv334WD');
+      expect(result).to.be.an('object');
```
krzyurb: I think it's not necessary to check if `result` is an `object`. In the next line, you are checking `members` field. If the `result` won't be an object it will fail when trying to access that field.

test/slack/users-ids-mapper.spec.js
kozikkam: Right now the service takes an array of user emails as parameter. I feel like it would be a bad practice to let it take either an array or a string.

test/slack/users-ids-mapper.spec.js
kozikkam: That won't work, `await usersIdMapper.requestUsers()` will throw an error, and tests won't pass. It is necessary to pass function as argument if the error is to be caught by `to.be.rejectedWith`.

src/slack/UsersIdMapper.service.js
kozikkam: As seen in the above if statement, params is necessary, since the result can be either for no cursor or with cursor.

kozikkam:
@krzyurb done

test/slack/users-ids-mapper.spec.js
krzyurb: The same situation [here](https://github.com/netguru/bob-the-bot/pull/76/files#diff-ac044187a83b684569cb741c380f2372R76)

test/slack/users-ids-mapper.spec.js
krzyurb: Consider doing the same [here](https://github.com/netguru/bob-the-bot/pull/76/files#diff-ac044187a83b684569cb741c380f2372R137) and [here](https://github.com/netguru/bob-the-bot/pull/76/files#diff-ac044187a83b684569cb741c380f2372R113)

test/slack/users-ids-mapper.spec.js
krzyurb: And how about using expect `.to.throw('Some error')`;?

test/book/users-ids-mapper.spec.js
```diff
@@ -0,0 +1,143 @@
...
+
+const UsersIdMapperService = require('../../src/book/UsersIdMapper.service');
+const SlackClientFactory = require('../../src/clients/SlackClientFactory');
+const errors = require('../../src/book/slack.errors');
+
+const client = SlackClientFactory();
+
+describe('Slack emails to ids mapper', () => {
+  const userEmails = ['super@malpka.pl'];
+  let usersIdMapper = new UsersIdMapperService(userEmails, client);
```
krzyurb: Consider moving it to `beforeEach`

src/book/UsersIdMapper.service.js
```diff
@@ -0,0 +1,51 @@
...
+      throw errors.NO_RESULTS;
+    }
+
+    return response;
+  }
+
+  async getMappedSlackUsers() {
+    const mappedUsers = [];
+    let response = await this.requestUsers();
+    for (let i = 0; i < 1; i += 1) {
```
krzyurb: I don't like that `for`. Try to execute that loop as long as `response.response_metadata.next_cursor` exists

kozikkam: It's not as simple as changing to a while loop. The block always has to run at least once, if changed to `do { } while()`, it could look like:
`    do {
      response.members.forEach((member) => {
        if (this.usersEmails.indexOf(member.profile.email) > -1) {
          mappedUsers.push({ [member.profile.email]: member.id });
        }
      });
      if (response.response_metadata.next_cursor) {
        response = await this.requestUsers(response.response_metadata.next_cursor);
      }
    } while(response.response_metadata.next_cursor)`

But then it doesn't work, because if 
`      if (response.response_metadata.next_cursor) {
        response = await this.requestUsers(response.response_metadata.next_cursor);
      }`

goes off, and response contains cursor = '', then the loop ends without executing forEach on response members.

kozikkam: @krzyurb

PR #77, opened 2018-10-15
BOOK-6 Fetch rooms resources

krzyurb:
Create a service to fetch rooms-resources data from google api

**Usage example**
```
const service = new RoomsService();
const rooms = await service.getRooms();
```

**Jira**
BOOK-6


src/book/rooms.service.js
```diff
@@ -0,0 +1,32 @@
+class RoomsService {
+  constructor(clientExtractor) {
+    this.clientExtractor = clientExtractor;
+  }
+
+  async googleResources() {
+    const client = await this.getClient();
+    const resources = await client.resources.calendars.list({ customer: 'my_customer' });
+    return resources.data.items;
```
Getriax: I think it would be nice to make sure that `.data` and `.data.items` are available. It is better to prevent the runtime error 🙂

src/book/rooms.service.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  async getRooms() {
+    let resources = await this.googleResources();
```
Getriax: Don't you think that it would be better to handle an exception that may occur in this line?

src/book/rooms.service.js
```diff
@@ -0,0 +1,32 @@
...
+    resources = resources.filter(resource => resource.resourceCategory === 'CONFERENCE_ROOM');
```
Getriax: Are you sure that resources are always an `array`?

krzyurb: yes

src/book/rooms.service.js
krzyurb: I think it will be better to catch an error in a plugin if it occurs here

src/book/rooms.service.js
krzyurb: According to docs, it always responds with `data.items`

src/book/rooms.service.js
```diff
@@ -0,0 +1,38 @@
+const errors = require('./book.errors');
```
kozikkam: Please change to `const { errors } = require('./book.errors');`
Otherwise it won't work.

PR #78, opened 2018-10-16
[BOOK-10] Remove event rooms

Getriax:
## Description

Create a service that removes all Netguru rooms from the selected event.

## How to use

    const service = ResourcesService(googleClientExtractor)
    const rooms = await roomsService.getRooms();

    await service.setRooms(rooms)
             .removeEventRooms(calendarId, eventId)


src/book/resources.service.js
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable class-methods-use-this */
+const { assertOrThrow } = require('../../utils');
+const { errors } = require('./book.errors');
+
+class ResourcesService {
+  constructor(clientExtractor) {
+    this.clientExtractor = clientExtractor;
+    this.rooms = null;
```
krzyurb: I think thank `rooms` should be set as `[]` to avoid errors when a developer will call `retreiveEmails` before `setRooms` (and how about passing also `rooms` as constructor param?).

Getriax:
@krzyurb fixes done! ⚒

PR #82, opened 2018-10-17
Book 9 book plugin

krzyurb:
**Task description**

Bot sends the message to the user that asks for confirmation. 

**Usage example**
aaaaa

**Jira**
BOOK-9


scripts/book.js
```diff
@@ -0,0 +1,19 @@
+// Description:
+//  Script to confirm rooms bookings
+//
```
Getriax:
```suggestion
//
// Configuration:
// CALENDAR_CLIENT_ID - calendar oauth client id
// CALENDAR_CLIENT_SECRET - calendar oauth secret
// CALENDAR_OAUTH_ROOM_ID - room of a person who can grant access to bots calendar
//
```

I think those properties are required here? 🙂

src/book/book.actions.js
```diff
@@ -0,0 +1,64 @@
...
+    this.slackClient = slackClient;
+    this.redis = redis;
+    autoBind(this);
+  }
+
+  async confirmBooking(res, payload) {
+    const { confirmed, event } = JSON.parse(payload.actions[0].value);
+    const timestamp = payload.original_message.ts;
+
+    const attachments = [];
```
Getriax: Why don't you use the `MessageBuilder` class here?

src/book/confirm.message.js
```diff
@@ -0,0 +1,34 @@
+const { MessageBuilder } = require('./../../utils');
+
+module.exports = (event) => {
+  const builder = new MessageBuilder();
+
+  builder.addField({ text: 'Can you confirm your meeting?' });
+  builder.addField({ callback_id: 'booking-confirmation' });
```
Getriax: It is possible to **chain** `addAttachment` and `addField` builder methods. You could even move the initial fields to the 4th line 🙂

test/book/book.actions.spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+  afterEach(() => {
+    sinon.restore();
+    room.destroy();
+  });
+
+  describe('confirm-booking callback', () => {
+    it('responds with 200 status when proper data was given', async () => {
+      sinon.stub(slackClient.chat, 'update').returns({ ok: true });
+      sinon.stub(redisClient, 'hget').callsFake(() => ('{ "ownerId": "QWE", "status": false }'));
```
Getriax: Why not use `JSON.stringify()` on a actual **JSON**? Also, as far as I'm concerned `redisClient.hget` returns a **promise**. I think it is important to make sure the return type is the same as in the original response.

```suggestion
      sinon.stub(redisClient, 'hget').resolves(JSON.stringify({ ownerId: 'QWE', status: false }))
```

test/book/book.actions.spec.js
```diff
@@ -0,0 +1,107 @@
...
+      const response = await request(room.robot.server)
+        .post(endpoint)
+        .send({ payload });
+
+      await tick(1);
+
+      expect(response.statusCode).to.equal(200);
+    });
+
+    it('responds with 500 status when wrong data was given', async () => {
```
Getriax: In this test there is an error `Cannot read property '0' of undefined` which happens at line 11.
   
      const { confirmed, event } = JSON.parse(payload.actions[0].value);

If this is a case please make sure that the `payload.actions` are not undefined.
https://github.com/netguru/bob-the-bot/pull/82/files#diff-49d751fe5206c0710abfff61e38ef1b0R11

Getriax: I think that you should stub the `updateEvent` method so it won't cause another error case.

test/book/book.actions.spec.js
Getriax: You are returning the plain value, where normally there is a promise returned, in multiple places.

*  https://github.com/netguru/bob-the-bot/pull/82/files#diff-acc9a22a174fd6abe6c1bcf7dfc4ef2cR46
* https://github.com/netguru/bob-the-bot/pull/82/files#diff-acc9a22a174fd6abe6c1bcf7dfc4ef2cR68
* https://github.com/netguru/bob-the-bot/pull/82/files#diff-acc9a22a174fd6abe6c1bcf7dfc4ef2cR69

krzyurb:
@Getriax fixes done

test/book/book.actions.spec.js
Getriax: I still believe that there should be a proper error thrown, maybe you could write a function that retrieves the `actions` from `payload` and if none specified you could `throw new Error('Insufficient request data given')`?

PR #84, opened 2018-10-19
Book 7 data fetcher

kozikkam:
**Task description**

Service that fetches all incoming events from google calendar, filters them - leaving only the ones that will happen in 20 minutes or less and have a Netguru rooms resources assigned and has assigned Bob.

Saves to Redis in the form of
```
eventsToConfirm: {
  [event.id]: {
    ownerId: String,
    name: String,
    url: String,
    confirmed: Boolean,
    requested: Boolean,
    start: String,
  },
}
```

**Usage example**
```
const eventsFetcher = new EventsFetcher(clientExtractor, roomResources, redisClient);
const filteredEvents = await eventsFetcher.getFilteredEvents();
```

Where

`clientExtractor` - google calendar api extractor

`roomResources` - list of room resources in the form of:
```  
{
    id: String,
    name: String,
    building: String,
    fullName: String,
    type: String,
    floor: String,
    email: String,
},
```

`redisClient` - redis client factory

**Jira**
BOOK-7


src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+  async saveToDatabase(events) {
+    let params;
+    let organizerMail;
+    let redisResult;
+    const added = await reduce(events, async (accumulator = [], event) => {
+      redisResult = await this.redis.hget(this.hash, event.id);
+      if (redisResult === null) {
+        accumulator.push(event);
+        organizerMail = this.usersMapper.getMappedSlackUsers([event.creator.email]);
+        params = {
```
krzyurb: How about moving it to separated function (e.g. `mapEvent`) which returns already prepared and stringified object for redis

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+          [event.id]: {
+            ownerId: organizerMail,
+            name: event.summary,
+            url: event.htmlLink,
+            confirmed: false,
+            requested: false,
+            start: event.start.dateTime,
```
krzyurb: Is it timestamp?

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+    return filtered;
+  }
+
+  async saveToDatabase(events) {
+    let params;
+    let organizerMail;
+    let redisResult;
+    const added = await reduce(events, async (accumulator = [], event) => {
+      redisResult = await this.redis.hget(this.hash, event.id);
+      if (redisResult === null) {
```
krzyurb: Add an empty line below

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+    });
+    
+    return filtered;
+  }
+
+  async saveToDatabase(events) {
+    let params;
+    let organizerMail;
+    let redisResult;
```
krzyurb: Add an empty line below

test/book/events-fetcher.service.spec.js
```diff
@@ -0,0 +1,210 @@
+const { expect } = require('chai');
+const sinon = require('sinon');
+
+const EventsFetcher = require('../../src/book/eventsFetcher.service');
+const redisClient = require('../../src/clients/RedisClientFactory');
+
+const mockedEmail = 'bob@email.com';
```
krzyurb: How about moving it to `.env.test`?

test/book/events-fetcher.service.spec.js
```diff
@@ -0,0 +1,210 @@
...
+  creator: {
+    id: 'creatorid1',
+    email: 'creator@email.com',
+  },
+  organizer: {
+    id: 'organizerid1',
+    email: 'organizer@email.com',
+  },
+  start: {
+    dateTime: (new Date(new Date().setMinutes(new Date().getMinutes() + 10))).toISOString(),
```
krzyurb: Try to use moment.js. This way is very unreadable

test/book/events-fetcher.service.spec.js
```diff
@@ -0,0 +1,210 @@
...
+    {
+      id: 'resourceid1',
+      email: roomResources[0].email,
+    },
+  ],
+}];
+
+let database = [];
+
+const redisStub = {
```
krzyurb: What do you think about this introducing [this one](https://www.npmjs.com/package/ioredis-mock) to the project?

test/book/events-fetcher.service.spec.js
```diff
@@ -0,0 +1,210 @@
...
+      later.setMinutes(later.getMinutes() + minsLater);
+
+      const minutes = fetcher.minutesToStart(later);
+
+      expect(minutes).to.equal(minsLater);
+    });
+
+    it('Should return a valid time difference for earlier', () => {
+      const minsEarlier = 20;
+      const earlier = new Date();
```
krzyurb: Consider using [.subtract](https://momentjs.com/docs/#/manipulating/subtract/)

test/book/events-fetcher.service.spec.js
```diff
@@ -0,0 +1,210 @@
...
+      expect(fetcher.roomEmails.length).to.equal(fetcher.roomResources.length);
+      expect(fetcher.roomEmails[0]).to.equal(fetcher.roomResources[0].email);
+    });
+  });
+
+  describe('minutesToStart', () => {
+    it('Should return a valid time difference for to come', () => {
+      const minsLater = 20;
+      const later = new Date();
+      later.setMinutes(later.getMinutes() + minsLater);
```
krzyurb: Consider using [.add](https://momentjs.com/docs/#/manipulating/add/)

test/book/events-fetcher.service.spec.js
```diff
@@ -0,0 +1,210 @@
...
+
+      expect(result).to.be.false;
+    });
+  });
+
+  describe('filterUpcomingWithResources', () => {
+    it('Should return filtered events', async () => {
+      const item2 = JSON.parse(JSON.stringify(items[0]));
+      item2.start = {
+        dateTime: (new Date(new Date().setMinutes(new Date().getMinutes() - 30))).toISOString()
```
krzyurb: Same as [here](https://github.com/netguru/bob-the-bot/pull/84/files#diff-5bc19e5f32494aaf4a025936be060dd5R43)

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+const slackClient = require('../clients/SlackClientFactory');
+
+class EventsFetcherService {
+  constructor(clientExtractor, roomResources, redisClient) {
+    this.clientExtractor = clientExtractor;
+    this.roomResources = roomResources;
+    this.redis = redisClient;
+    this.time = new Date();
+    this.roomEmails = [];
+    this.usersMapper = new UsersMapper(slackClient());
```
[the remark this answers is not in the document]
kozikkam: In that case I will pass usersMapper as argument, and there will be no slackClient (it was there only for initialization of UsersMapper).

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  setRoomEmails() {
+    this.roomResources.forEach((room) => {
+      this.roomEmails.push(room.email);
+    });
+  }
+
+  minutesToStart(date) {
+    this.time = new Date();
```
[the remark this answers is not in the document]
kozikkam: Changed this.time to this.now (as per previous request). It's required to calculate how long till event start.

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+    const attendeesArray = Array.from(attendees);
+    return attendeesArray.some((attendee) => {
+      return this.roomEmails.indexOf(attendee.email) > -1;
+    });
+  }
+
+  async filterUpcomingWithResources(events) {
+    const filtered = [];
+    let minutesToStart;
+    events.forEach((event) => {
```
krzyurb: Please use `.filter` instead of `.forEach`.

src/book/eventsFetcher.service.js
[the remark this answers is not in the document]
krzyurb: So, why you need `this.time` and `this.now`? Both object will point to almost the same date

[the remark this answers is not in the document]
kozikkam: Valid point, was useless

test/book/events-fetcher.service.spec.js
kozikkam: It really comes down to
`later = moment(later).subtract(20, "minutes").toDate()`
or
`later.setMinutes(later.getMinutes() + minsLater);`

I wouldn't say it's easier to read.

test/book/events-fetcher.service.spec.js
kozikkam: Same as above

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+      calendarId: process.env.BOT_EMAIL,
+    };
+    const client = await this.getClient();
+    const res = await client.events.list(params);
+
+    return res.data.items;
+  }
+
+  containsRoom(attendees) {
+    const attendeesArray = Array.from(attendees);
```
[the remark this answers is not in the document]
kozikkam: Otherwise there's an error
`TypeError: attendees.some is not a function`

src/book/eventsFetcher.service.js
kozikkam: According to google api documentation
```
"start": {
  "date": date,
  "dateTime": datetime,
  "timeZone": string
},
```

```
The time, as a combined date-time value (formatted according to RFC3339). A time zone offset is required unless a time zone is explicitly specified in timeZone.
```

I'll change variable name to dateTime.

test/book/events-fetcher.service.spec.js
kozikkam: If we did that with every mocked variable, each new person would have to do a ton of work (or get prepared .env.test) to run tests. I think it's pointless.

kozikkam:
@krzyurb Done

src/book/eventsFetcher.service.js
```diff
@@ -21,10 +20,10 @@ class EventsFetcherService {
...
 
   minutesToStart(date) {
-    this.time = new Date();
-    const localDate = new Date(date);
-    const timeDiff = (localDate - this.time);
-    const diffMins = Math.round(((timeDiff % 86400000) % 3600000) / 60000);
+    const now = moment(new Date());
+    const end = moment(date);
+    const duration = moment.duration(end.diff(now));
+    const diffMins = Math.round(duration.asMinutes());
```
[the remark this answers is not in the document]
kozikkam: Everything can be done in one line if you try hard enough. It's about readability. You miss Math.round() here, and also omit definition of `end`. It could be done as:
```
const end = moment(date);
return Math.round(moment(end).diff(moment(), 'minutes'));
```
I don't really see the upside here. One liner bonus:
```
return Math.round(moment(moment(date)).diff(moment(), 'minutes'));
```
I guess I'll change it to two liner.

test/book/events-fetcher.service.spec.js
krzyurb: We have a [.env.test.example](https://github.com/netguru/bob-the-bot/blob/master/.env.test.example) file. I think it might be very useful in future

src/book/eventsFetcher.service.js
```diff
@@ -23,12 +22,8 @@ class EventsFetcherService {
   }
 
   minutesToStart(date) {
-    const now = moment(new Date());
     const end = moment(date);
-    const duration = moment.duration(end.diff(now));
-    const diffMins = Math.round(duration.asMinutes());
-
-    return diffMins;
+    return Math.round(moment(end).diff(moment(), 'minutes'));
```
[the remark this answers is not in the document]
kozikkam: Tests have shown that if there is exactly 20 minutes difference between two dates, then without Math.round the result is 19 minutes.

krzyurb:
@kozikkam Please check filing tests

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
...
+    const end = moment(date);
+    const duration = moment.duration(end.diff(now));
+    const diffMins = Math.round(duration.asMinutes());
+    
+    return diffMins;
+  }
+
+  async getAllEvents() {
+    const client = await this.getClient();
+    const res = await client.events.list({
```
Getriax: You don't have to `filter` those events by **yourself**, you can just call the API with proper parameters.
Consider this:

       const res = await client.events.list({
          calendarId: process.env.BOT_EMAIL,
          timeMin: now,
          timeMax: 20 minutes ahead,
          singleEvents: true
       });

Please refer to https://developers.google.com/apis-explorer/#search/calendar/calendar/v3/calendar.events.list

Consider removing the filtering logic from the code, and adjust the code to API possibilities.

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable no-await-in-loop */
```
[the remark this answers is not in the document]
kozikkam: done

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
...
+const { reduce } = require('asyncro');
+const moment = require('moment');
+const { redisEventsKey } = require('../../config/book.config.json');
+const assertOrThrow = require('../../utils/AssertOrThrow');
```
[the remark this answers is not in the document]
kozikkam: done

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
...
+const errors = require('./book.errors');
```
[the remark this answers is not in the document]
kozikkam: done, good catch

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
...
+    this.redis = redisClient;
+    this.usersMapper = usersMapper;
+  }
+
+  getRoomEmails() {
+    const roomEmails = [];
+    this.roomResources.forEach((room) => {
+      roomEmails.push(room.email);
+    });
+    return roomEmails;
```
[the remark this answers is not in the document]
kozikkam: You mean new line, right? Added

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
...
+    
+    return diffMins;
+  }
+
+  async getAllEvents() {
+    const client = await this.getClient();
+    const res = await client.events.list({
+      calendarId: process.env.BOT_EMAIL,
+    });
+    assertOrThrow(res, errors.GOOGLE_API_ERROR);
```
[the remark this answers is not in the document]
kozikkam: I have mixed feelings about this, but added.

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable no-await-in-loop */
+const { reduce } = require('asyncro');
+const moment = require('moment');
+const { redisEventsKey } = require('../../config/book.config.json');
```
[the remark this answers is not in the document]
kozikkam: Yeah, saw that as well, added.

src/book/eventsFetcher.service.js
kozikkam: As it turns out, google api doesn't work as intended. However such way of requesting does improve performance significantly. Thx.

kozikkam:
@krzyurb @Getriax Done

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,114 @@
...
+    const diffMins = Math.round(duration.asMinutes());
+
+    return diffMins;
+  }
+
+  async getAllEvents() {
+    const client = await this.getClient();
+    const res = await client.events.list({
+      calendarId: process.env.BOT_EMAIL,
+      timeMin: moment(),
```
Getriax: Shouldn't it be `moemnt().toDate()`?
```suggestion
      timeMin: moment().toDate(),
```

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  async getAllEvents() {
+    const params = {
+      calendarId: process.env.BOT_EMAIL,
+    };
+    const client = await this.getClient();
+    const res = await client.events.list(params);
+
+    return res.data.items;
```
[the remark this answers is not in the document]
Getriax: Have you added the `try...catch`?

src/book/eventsFetcher.service.js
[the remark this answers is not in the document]
Getriax: Have you changed that to two-liner?

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,114 @@
...
+      return (minutesToStart > 0 && this.containsRoom(event.attendees, roomEmails));
+    });
+
+    return filtered;
+  }
+
+  async mapEvent(event) {
+    let res;
+    try {
+      res = await this.usersMapper.getMappedSlackUsers([event.creator.email]);
```
Getriax: Can you gather all the emails at first and then call the `.getMappedSlackUsers` once? Remember that for each usage of this method there are `1...n` API calls. Depends on pagination.  Having all the emails, and call this method once makes this process more efficient.

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,114 @@
...
+        requested: false,
+        dateTime: event.start.dateTime,
+      },
+    });
+  }
+
+  async saveToDatabase(events) {
+    let redisResult;
+
+    const added = await reduce(events, async (accumulator = [], event) => {
```
Getriax: `array.prototype.reduce` takes three params, according to https://www.npmjs.com/package/asyncro.

```
await reduce(
    ['/foo', '/bar', '/baz'],
    async (accumulator, value) => {
        accumulator[v] = await fetch(value);
        return accumulator;
    },
    {}
);
```

Please remove the `= []` from `accumulator = []` and add the `[]` as the third parameter.

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,114 @@
...
+      redisResult = await this.redis.hget(redisEventsKey, event.id);
```
Getriax: You can remove the `let redisResult` and make it `const` here.
```suggestion
      const redisResult = await this.redis.hget(redisEventsKey, event.id);
```

Also you can just return the result of the reduce, no need of `const added; return added;`

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,114 @@
...
+  constructor(clientExtractor, roomResources, redisClient, usersMapper) {
+    this.clientExtractor = clientExtractor;
+    this.roomResources = roomResources;
+    this.redis = redisClient;
+    this.usersMapper = usersMapper;
+  }
+
+  getRoomEmails() {
+    const roomEmails = [];
+    this.roomResources.forEach((room) => {
```
Getriax: Please use `map` here. Also, what are those  `this.roomResources`? Should I retrieve the rooms from google and pass them here?

src/book/eventsFetcher.service.js
kozikkam: Good catch

src/book/eventsFetcher.service.js
kozikkam: Good catch

src/book/eventsFetcher.service.js
kozikkam: changed

src/book/eventsFetcher.service.js
kozikkam: changed

src/book/eventsFetcher.service.js
kozikkam: I don't think changing roomResources is a good idea.
Second question: yes, pass them to constructor.

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,130 @@
...
+      return acc;
+    }, []);
+
+    try {
+      mappedSlackUsers = await this.usersMapper.getMappedSlackUsers(userEmails);
+    } catch (err) {
+      throw err;
+    }
+
+    events.forEach((event) => {
```
Getriax: Please use `map` here

kozikkam: Done

src/book/eventsFetcher.service.js
```diff
@@ -0,0 +1,130 @@
...
+    return diffMins;
+  }
+
+  containsRoom(attendees, roomEmails) {
+    return attendees.some(attendee => roomEmails.includes(attendee.email));
+  }
+
+  stringifyEvent(event) {
+    return JSON.stringify({
+      [event.creator.email]: {
```
Getriax: We agreed that the object in the database should look like:

    eventsToConfirm: {
        EventId: {
	      ownerId, // slack user ID
	      ownerEmail: // owner Email
	      name: // event name
	      url: // event url
	      confirmed: false,
	      requested: false
	      start // moment rozpoczęcia eventu
      }
    }

Why are you making this like: 

    eventsToConfirm: {
        EventId: {
		     ownerEmail: {
			    ownerId, // slack user ID
		      name: // event name
		      url: // event url
		      confirmed: false,
		      requested: false
		      start // moment rozpoczęcia eventu
	      }
	    }
	}

Please correct it

src/book/eventsFetcher.service.js
kozikkam: changed

src/book/eventsFetcher.service.js
kozikkam: Fixed

kozikkam:
@Getriax Done

src/book/eventsFetcher.service.js
kozikkam: Also, regarding `this.roomResources`, you could have looked it up here https://github.com/netguru/bob-the-bot/pull/84#issue-224345886 (pr's description).

kozikkam:
@Getriax done

PR #85, opened 2018-10-23
Create BD support plugin

krzyurb:
## How it works
Dialog
Link


src/jira/jira.api.js
```diff
@@ -2,7 +2,7 @@
 const axios = require('axios');
 const { errors } = require('./jira.errors');
 const { assertOrThrow } = require('../../utils');
-const { query } = require('../../config/jira.config.json');
+const { query, bdProject } = require('../../config/jira.config.json');
```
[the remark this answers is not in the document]
Getriax: I think that step is good but unnecessary here. Because moving forward there should be a possibility to create tickets not only for bdSupport. I think that in this case, we would make another change. 

Also what about `confluence` which is also using the `AtlassianClientFactory`? It would be quite odd seeing the `jira.api.js` there whilst the `confluence.api.js` is still in the `confluence folder`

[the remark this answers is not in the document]
krzyurb: Than both of files could be moved to `clients/atlassian` in future

src/bdSupport/bdSupport.actions.js
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable class-methods-use-this */
+const autoBind = require('auto-bind');
+
+const { MessageRenderer, DialogBuilder } = require('../../utils');
+
+class DbSupportActions {
```
kozikkam: Why are you using both `DB` and `BD` phrases? Please fix naming.

src/bdSupport/bdSupport.actions.js
```diff
@@ -0,0 +1,114 @@
...
+  constructor(jiraClient, slackClient) {
+    this.jiraClient = jiraClient;
+    this.slackClient = slackClient;
+
+    autoBind(this);
+  }
+
+  retreiveActionUser(actions) {
```
kozikkam: Please fix function name to 'retrieveActionUser`

scripts/bdSupport.js
```diff
@@ -0,0 +1,31 @@
...
+const BdSupportActions = require('../src/bdSupport/bdSupport.actions');
+const AtlassianClientFactory = require('../src/clients/AtlassianClientFactory');
+const SlackClientFactory = require('../src/clients/SlackClientFactory');
+const { SlackActionsMultiplexer, dependenciesLocator } = require('../utils');
+
+module.exports = (robot) => {
+  dependenciesLocator.set('slackActionsMultiplexer', new SlackActionsMultiplexer(robot));
+
+  const { jiraAPI } = AtlassianClientFactory();
+  const actions = new BdSupportActions(jiraAPI, SlackClientFactory());
```
kozikkam: Why are you destructuring jiraApi, but calling `SlackClientFactory`?

test/bdSupport/bdSupport.actions.spec.js
```diff
@@ -0,0 +1,329 @@
+const Helper = require('hubot-test-helper');
+const sinon = require('sinon');
+const { expect } = require('chai');
+const request = require('supertest');
+
+const moment = require('moment');
```
kozikkam: Why split `moment` from the rest?

Getriax: `hubot-test-helper`, `sinon`, `chai`, `supertest` are *testing* libraries whilst `moment` is the *date* library.

PR #88, opened 2018-10-31
BD support fixes

krzyurb:


src/bdSupport/bdSupport.actions.js
```diff
@@ -112,6 +155,26 @@ class BDSupportActions {
...
+        ts: moment().format('X'),
+      }],
+    });
+
+    const permalinkResponse = await this.slackClient.chat.getPermalink({
+      channel: process.env.BDTECH_ROOM_ID,
+      message_ts: messageResponse.ts,
+    });
+
+    return { message: messageResponse, permalink: permalinkResponse.permalink };
```
Getriax: You are not using the `message` property in the code. Why do you need to return it?

src/jira/jira.api.js
```diff
@@ -37,6 +37,13 @@ class JiraAPI {
     return result;
   }
 
+  async createIssueComment(id, commentBody) {
+    const data = { body: commentBody };
+    const result = await this.jiraRequest(`/issue/${id}/comment`, { data }, 'post');
+
+    return result;
```
Getriax: You are not using return value in code, why return it then?

templates/bdsupport/description.message
```diff
@@ -2,11 +2,13 @@
 {{{description}}}
 
 -----------------------
-Category: {{{category}}}
+
 OPP Link: {{{oppLink}}}
```
Getriax: What if `OPP Link` is not specified? Do you want to post it anyway?

test/bdSupport/bdSupport.actions.spec.js
```diff
@@ -147,7 +149,125 @@ describe('bdSupprot actions test suite', () => {
...
+        .post(actionsEndpoint)
+        .send({ payload: JSON.stringify(payload) });
+
+
+      await tick(1);
+
+      expect(response.statusCode).to.equal(200);
+    });
+
+    it('Should respond with status 500 if cannot create dialog', async () => {
```
Getriax: This test case throws an error `ERROR SyntaxError: Unexpected token u in JSON at position 0` this does not look nice, could you fix it? 🙂 If you are expecting this error, could you catch it and check if the proper error was thrown?

src/bdSupport/bdSupport.actions.js
```diff
@@ -21,6 +21,28 @@ class BDSupportActions {
...
   }
 
+  async slashCommandDialog(res, payload) {
+    const user = { id: payload.user_id, name: payload.user_name };
+
+    const userData = await this.slackClient.users.profile.get({ user: user.id });
+    user.email = userData.ok ? userData.profile.email : '';
+
+    const callBackId = 'create-slash-db-support-ticket';
+    const dialog = new DialogBuilder(
```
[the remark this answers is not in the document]
krzyurb: Done

templates/bdsupport/description.message
krzyurb: Yes

krzyurb:
@Getriax Fixes done

PR #90, opened 2018-11-01
BOOK-20 MQTT integration

krzyurb:
Implementation of MQTT protocol in bot. All messages sent to bot should match schema:
```
{
  action: "action-name",
  data: {
    exampleText: "abcde",
    ...
  }
}
```
[aws-iot-device-sdk-js](url) library was used to handle aws iot connection.

**Usage example**

Listening to messages:
```
  mqtt.onAction('topic-name', 'action-name', (data) => {

  });
```

Sending messages:
```
mqtt.sendMessage('topic-name', { exampleData: 123 });
```

**JIRA**
BOOK-20


test/book/book.processes.spec.js
```diff
@@ -25,12 +26,17 @@ const redisClient = {
   hset: () => Promise.resolve(),
 };
 
+class Device extends EventsEmitter { };
```
kozikkam: In theory you shouldn't put semicolons after class definitions.

test/book/book.processes.spec.js
```diff
@@ -25,12 +26,17 @@ const redisClient = {
...
+const device = new Device();
```
kozikkam: What's the point of this? Why not do `const device = new EventsEmitter()`?

test/book/book.processes.spec.js
```diff
@@ -47,6 +53,7 @@ describe('Book processes', () => {
   afterEach(() => {
     sinon.restore();
     room.destroy();
+    dependenciesLocator.forceSet('mqtt', null);
```
kozikkam: Setting to null afterEach is reduntand, since you do
`dependenciesLocator.forceSet('mqtt', new Mqtt(device));`
beforeEach.

test/book/book.spec.js
```diff
@@ -34,13 +35,17 @@ const bullClient = {
 };
 
 const tick = ms => new Promise(resolve => setTimeout(resolve, ms));
+class Device extends EventsEmitter { };
+const device = new Device();
```
kozikkam: Duplicate of what above.

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
+const EventsEmitter = require('events');
+const chai = require('chai');
+const sinon = require('sinon');
+
+const { expect } = chai;
+const { Mqtt } = require('./../../utils');
```
kozikkam: Duplicate as above.

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+chai.use(require('sinon-chai'));
+
+class Device extends EventsEmitter {};
+const device = new Device();
```
kozikkam: Duplicate of above.

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
+const EventsEmitter = require('events');
+const chai = require('chai');
+const sinon = require('sinon');
+
+const { expect } = chai;
+const { Mqtt } = require('./../../utils');
+
+chai.use(require('sinon-chai'));
+
+class Device extends EventsEmitter {};
```
Getriax:
```suggestion
class Device extends EventsEmitter {
   publish(topic, payload) => {
     device.emit('message', topic, payload);
   };
};
```

I think that by doing this you are able to test the `mqtt.sendMessage` method, which is not tested right now. What do you think about that? 🙂

Maybe even create a DeviceFactory?

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+    mqtt.onAction('topic', 'first', firstFunction);
+    mqtt.onAction('topic', 'second', secondFunction);
+
+    const payload = { action: 'third' };
+
+    sendMessage('topic', payload);
+
+    expect(firstFunction).to.not.have.been.calledOnceWith();
+    expect(secondFunction).to.not.have.been.calledOnceWith();
```
kozikkam: You might want to remove half of this function's code. It is unnecessary to register two listeners, one is enough.

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    sendMessage('topic', payload);
+
+    expect(firstFunction).to.be.calledOnceWith(payload.data);
+  });
+
+  it('should not throw an error when action has been duplicated', async () => {
+    mqtt.onAction('topic', 'first', () => {});
+
+    expect(() => {
+      mqtt.onAction('topic', 'first', () => { });
```
kozikkam: You do `() => {}` everywhere, why do `() => { }` here.

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    }).to.throw(Error, 'Mqtt action duplication');
```
kozikkam: Consider splitting that to
```
    mqtt.onAction('topic', 'first', () => {});

    const actionDuplicate = () => mqtt.onAction('topic', 'first', () => {});

    expect(actionDuplicate).to.throw(Error, 'Mqtt action duplication');
```

test/utils/mqtt.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+    const payload = { action: 'first' };
+
+    sendMessage('topic', payload);
+
+    expect(firstFunction).to.have.been.calledOnceWith();
+    expect(secondFunction).not.have.been.calledOnceWith();
+  });
+
+  it('should not execute any action if not defined', async () => {
```
Getriax: You are actually checking that case in the function above, https://github.com/netguru/bob-the-bot/pull/90/files#diff-89398a50bdaa87ae44ca697f617bbf2fR51 🙂

utils/Mqtt/src/Mqtt.js
```diff
@@ -0,0 +1,44 @@
...
+
+    this.listen();
+  }
+
+  static fullActionName(topic, actionName) {
+    return `${topic}/${actionName}`;
+  }
+
+  isActionDefined(action) {
+    return action && this.actions[action] !== undefined;
```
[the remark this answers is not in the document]
kozikkam: It will, but I stand with Krzysztof here, it's better to check explicitly.
But then again `this.actions[action]` can only be either undefined or a function, so stuff like `this.actions[action] -> true` won't be happening

test/utils/mqtt.spec.js
krzyurb: You are right. I'm removing this case.

test/utils/mqtt.spec.js
krzyurb: Greate idea! Added `publish` method, and tested `sendMessage`. 
But I don't see a point for now to create DeviceFactory.

test/book/book.processes.spec.js
krzyurb: Right, I forgot to remove it!

src/clients/AwsIotClientFactory.js
```diff
@@ -0,0 +1,23 @@
+const awsIot = require('aws-iot-device-sdk');
+const path = require('path');
+
+const { topics } = require('./../../config/iot.config.json');
```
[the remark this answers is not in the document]
krzyurb: I'm used to add `./` always at the beginning because when you make `require('./myModule')` it takes imports file from same folder, and when `require('myModule')` it imports from `node_module` ;). But you are right - I don't need it here.

krzyurb:
@Getriax @kozikkam Fixed. Thanks for the great suggestions

PR #93, opened 2018-11-06
[Improve] BD Support Slash Command

Getriax:
## Changes

* Removed the default command, only slash command remains.
* Jira ticket has now info about channel name and its type
* Links only to public channels


src/bdSupport/bdSupport.actions.js
```diff
@@ -47,125 +41,61 @@ class BDSupportActions {
...
-      const { key } = await this.jiraClient.createIssue({ ...submission });
-
-      const message = this.buildMessage(user, submission, key, description);
-      await new SlackActionRequest(responseUrl, false).request(message);
-
-      const permalink = await this.postSummaryMessage(user, submission, key, description);
-      this.jiraClient.createIssueComment(key, permalink);
-
-      return res.status(200).end();
+      return response.ok;
```
krzyurb: Do I understand it proper that you check is conversation public if the response is ok? 
Why you are not using [channels.list](https://api.slack.com/methods/channels.list) from slack api? It returns data like:

```
        "is_shared": false,
        "is_org_shared": false,
        "is_member": true,
        "is_private": false,
        "is_mpim": false,
```

Getriax: Retrieving all the available channels just to check one at least, in my opinion, isn't the best solution 🙂Hove ever you make the good point here because there should be the better option for checking that. I tried to implement [conversations.info](https://api.slack.com/methods/conversations.info) unfortunately it is not working as it supposed to, giving me an error while trying to get the information about private channel `channel_not_found`, having all the scopes idk what is wrong with the API [test branch](https://github.com/netguru/bob-the-bot/compare/improve/bd...test/bd) if you don't mind check it yourself, maybe I'm missing something. Until then `channels.info` is the best shot here 🤔

krzyurb: Ok, I remember now I had almost the same problem and I used the same solution

PR #96, opened 2018-11-07
[Fix] BOOK-29 Google Authenticator

Getriax:
## What was wrong

The emitted tokens do not include the `refresh_token`, only the first emit includes that property. Currently, the token is overwritten by the one emitted.

## Solution

Fixed this problem by assigning the new properties to the existing token, instead of overwriting it.

## Additional changes

To send the information about the authentication the function that handles the new tokens now also checks if the given one wasn't emitted by the user action.


test/clients/google.client.spec.js
```diff
@@ -45,9 +45,33 @@ describe('Google extractor client test suite', () => {
...
 
-  it('Should check that the room has the auth url', async () => {
+  it('Should send the auth url if token is not available on init', async () => {
     sinon.stub(redisClient, 'get').rejects();
     sinon.stub(googleClient.oAuthClient, 'getToken').callsFake(() => { });
+    sinon.stub(googleClient, 'robot').value(null);
+
+    await googleClientExtractor(room.robot);
+
+    tick(1);
```
krzyurb: Should be: `await tick(1);`

src/clients/GoogleClientExtractorFactory.js
```diff
@@ -77,12 +78,16 @@ class GoogleClient {
     await this.oAuthClient.getTokenInfo(token.access_token);
   }
 
+  throwRefreshTokenError() {
+    const renderer = new MessageRenderer('isfree/auth');
```
krzyurb: How about moving it from `isfree/auth`? It's not used only by `isFree`

src/clients/GoogleClientExtractorFactory.js
```diff
@@ -120,20 +126,40 @@ class GoogleClient {
...
+
+        Object.assign(this.token, token);
+
+        this.setCredentials(this.token);
       } catch (error) {
-        this.robot.logger.error(error.message);
+        return this.robot.logger.error(error);
       }
+
+      return this.sendSuccess();
```
krzyurb: No need to put `return` here if `sendSucces()` is some process

src/clients/GoogleClientExtractorFactory.js
```diff
@@ -120,20 +126,40 @@ class GoogleClient {
...
+  sendSuccess() {
+    if (this.requestPerformed) {
+      this.messageMasterRoom('Authorization succeeded');
+      this.requestPerformed = false;
+    }
+  }
+
   initRefreshTokenEvent() {
-    this.oAuthClient.on('tokens', (token) => {
+    this.oAuthClient.on('tokens', async (token) => {
```
krzyurb: Why this one is `async`?

Getriax: There should not be async, propably was verifying token there and forgot to remove that 😅

src/clients/GoogleClientExtractorFactory.js
```diff
@@ -120,20 +126,40 @@ class GoogleClient {
...
   initRefreshTokenEvent() {
     this.oAuthClient.on('tokens', (token) => {
       try {
-        this.setCredentials(token);
-        this.messageMasterRoom('Authorization succeeded');
+        this.checkRefreshToken(token);
+
+        Object.assign(this.token, token);
+
+        this.setCredentials(this.token);
```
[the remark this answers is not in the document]
Getriax: Actually, as you can see in the PR description I cannot override the token, rather call `Object.assign` on it. However, I've removed the `this.token = token` from `setCredentials`

test/clients/google.client.spec.js
```diff
@@ -45,9 +45,10 @@ describe('Google extractor client test suite', () => {
     room.destroy();
   });
 
-  it('Should check that the room has the auth url', async () => {
+  it('Should send the auth url if token is not available on init', async () => {
     sinon.stub(redisClient, 'get').rejects();
     sinon.stub(googleClient.oAuthClient, 'getToken').callsFake(() => { });
+    sinon.stub(googleClient, 'robot').value(null);
```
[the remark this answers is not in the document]
Getriax: It does because the robot cannot be assigned for `GoogleClientExtractorFactory` to initialize the post route.

test/clients/google.client.spec.js
```diff
@@ -121,24 +144,47 @@ describe('Google extractor client test suite', () => {
...
+      ['hubot', renderer.render()],
+    ];
+
+    expect(room.messages).to.eql(expected);
+  });
+
+  it('Should  send an information about missing refresh token, on first authentication', async () => {
+    sinon.stub(googleClient, 'robot').value(room.robot);
+
+    sinon.stub(googleClient, 'token').value({});
```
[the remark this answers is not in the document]
Getriax: `googleClient ` is a singleton, so `token` value may not be `{}`

Getriax:
@kozikkam ⚒  done

test/clients/google.client.spec.js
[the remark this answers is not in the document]
kozikkam: How come nothing breaks after removing then?

PR #101, opened 2018-11-12
Book 21,28

kozikkam:
Implemented mqtt handling with use of BOOK-20 mqtt protocol.

Bot now does two things. He listens for emitted events on `rooms-general`. And uses `resourcesManager.handleResponse(response)` to handle that event.
He also emits events to each room channel with following data structure:
```
{
  id: 'someUniqueId'
}
```
Where uuid is an id which references to redis entry, storing listened event data. That data is then retrieved when response comes.

It is now required to set `CONFIRMED_EXPIRE_SECONDS` in .env. That's the value of how many seconds listened event data is going to be stored in redis database. It has to be above timeout set on device.

**Usage example**

Listening to events
```
  mqtt.onAction('rooms-general', 'get-status', (data) => {
    this.bookService.manageMqttResponse(data);
  });
```

Publishing events
```
await this.mqttHandler.prepareAndPublish(toPublishEventsFilter);
```

**Jira**
BOOK-21 and BOOK-28


src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,40 @@
...
+    const eventKeys = Object.keys(events);
+
+    await map(eventKeys, async (eventKey) => {
+      const event = events[eventKey];
+      const rooms = await this.bookService.getEventRoomAttendees(event);
+
+      await map(rooms, async (room) => {
+        const id = await this.saveForLater(event, room);
+
+        this.mqtt.sendMessage(room.id, { uuid: id });
```
Getriax: Where does `room.id` come from? Because in my opinion there is not an `id` property under the `room`.

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,40 @@
...
+
+        this.eventsDAO.update(eventKey, { requestedListener: true });
+      });
+    });
+  }
+
+  async saveForLater(event, room) {
+    const body = {
+      eventId: event.id,
+      roomId: room.id,
```
Getriax: Same here, you sure about that `room.id`?

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,40 @@
...
+    this.mqtt = mqtt;
+    this.eventsDAO = eventsDAO;
+  }
+
+  async prepareAndPublish(events) {
+    const eventKeys = Object.keys(events);
+
+    await map(eventKeys, async (eventKey) => {
+      const event = events[eventKey];
+      const rooms = await this.bookService.getEventRoomAttendees(event);
```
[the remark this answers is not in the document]
Getriax: Also you should not be using the `index.js` file here, it should be another way around.

src/book/services/index.js
```diff
@@ -21,6 +22,31 @@ class BookService {
...
     this.jobsService = new JobsService(this.processes, eventsDAO);
+
+    this.resourcesManager = new ResourcesManager(this.resourcesService, eventsDAO);
+  }
+
+  async getEventRoomAttendees(event) {
+    try {
+      const rooms = await this.roomsService.getRooms();
+
+      return this.resourcesService.filterNotRooms(
```
Getriax: I would like to notice that the saved `event` does not have the `attendees` property.

It is saved in here:
https://github.com/netguru/bob-the-bot/blob/project/book/src/book/services/eventsFetcher.js#L37

The event that you are receiving here comes from `eventsDAO.getGrouped` which means that you should get an error here because filter looks like:

     filterNotRooms(roomEmails, { attendees }) {
       return attendees.filter(attendee => roomEmails.indexOf(attendee.email) !== -1);
     }

This way you should get an error `cannot read property filter of undefined` that also mens that you are whedther not testing this properly or not tasting at all.

src/book/index.js
```diff
@@ -20,6 +20,6 @@ module.exports = (robot) => {
   });
 
   mqtt.onAction('rooms-general', 'get-status', (data) => {
-    // BOOK-21 triggers here
+    this.bookService.manageMqttResponse(data);
```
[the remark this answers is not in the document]
kozikkam: Yeaah it's not being tested right now, I am assuming that `mqtt.onAction` has been unit tested on it's own, and code referenced here has it's own unit tests as well.

src/book/services/index.js
kozikkam: I am testing it, but it's about assuming what is being returned by `eventsDAO.getGrouped`, hence the error.

src/book/book.controller.js
```diff
@@ -10,8 +11,11 @@ class BookController {
     const {
       expiredEvents,
       needConfirmationEvents,
+      toPublishEventsFilter,
     } = await this.eventsDao.getGrouped();
 
+    await this.mqttHandler.prepareAndPublish(toPublishEventsFilter);
```
Getriax: You should not be invoking this function here. In my opinion `bull` process is required here.

`performCronJob ` is invoked every 10 minutes, taking the events 20 minutes ahead. So by sending the event to the device which has the fixed timeout for the response, you could end up in the situation that the device will send `false` before the event even begins. 

Please consider creating the bull process that will invoke that function when the event starts 🙂

src/book/services/index.js
Getriax: If you weren't sure you should have asked, but evnetsDAO suggests that the result comes from the database 🙂

src/book/services/mqttHandler.js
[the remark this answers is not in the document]
kozikkam: > The name `getEventRoomAttendees` suggests that you are receiving the **attendees array** here. However, you are assigning the result to the `rooms` which is really confusing. Please decide whether is it `rooms` or `attendees` 🙂

It only depends on the way you look at it, for me it means get Event's Attendees that are Rooms. `cont rooms` is called appropriately right now. For more clarity it could be called `roomsAttendees`, but in reality, those are objects of `Room` structure, so I think it should stay that way.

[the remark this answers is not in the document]
kozikkam: > https://github.com/netguru/bob-the-bot/pull/101/files#diff-06edef1aead55c61a94a0f4ec19a07fbR33
> 
> You are filtering for `NotRooms` and returning it here to the `rooms` variable

Yes. I want to filter out attendees that aren't rooms.

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,40 @@
...
+    });
+  }
+
+  async saveForLater(event, room) {
+    const body = {
+      eventId: event.id,
+      roomId: room.id,
+    };
+    const id = uuid();
+    await this.eventsDAO.setExpire(id, body);
```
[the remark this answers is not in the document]
kozikkam: Apparently it didn't occur to me that there is such unique field already present

src/book/services/mqttHandler.js
kozikkam: `src/book/services/rooms` `getRooms()`

src/book/book.controller.js
kozikkam: I cannot end in a situation where event is requested twice, because events are modified to contain property `requestedListener = true`. Refer to `mqttHandler` and `filters/toPublish.events.js`.

src/book/services/index.js
kozikkam: Yeah nothing to ask for really, I wrote the whole skeleton of `eventsFetcher`. It just didn't occur to me that we would for some reason not save attendees. It's a quick fix anyway.

src/book/book.controller.js
Getriax: I am aware of `requestedListener = true` and this is not the case here. Please follow my example once more:

* event starts in 19 minutes so it is retrieved and sent to the device
* device has fixed property TIMEOUT, let's assume it is 10 minutes
* So device immediately listens for 10 minutes but the event has not even begun.
* 10 minutes later the device sends that there wasn't anyone in the room, so the room booking is removed
* the event starts 9 minutes later without a room

Can you see my point now?

kozikkam: > event starts in 19 minutes so it is retrieved and sent to the device
Look at `/filters/toPublish.events.js`.

You can see there:
```
    return moment()
      .isSameOrAfter(event.start)
      && event.confirmed
      && (!event.requestedListener
        || event.requestedListener === false);
```

So supposedly there can't be an event sent to a device, if it hasn't yet started.

src/book/index.js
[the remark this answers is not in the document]
Getriax: So maybe an integration test would be the resolution here?

src/book/events.dao.js
```diff
@@ -19,6 +20,18 @@ class EventsDAO {
...
 
+  parseOrThrow(event, id) {
+    if (!event) {
+      throw new Error(`Event with ID ${id} not found`);
+    }
+
+    return JSON.parse(event);
+  }
+
+  async setExpire(id, eventBody, expire = process.env.CONFIRMED_EXPIRE_SECONDS) {
```
Getriax: You are creating a global redis entry, the expire should be resolved within the event body which is already present in the database. Consider using `evnetsDAO.update`.

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,44 @@
...
+
+        this.mqtt.sendMessage(room.id, { id });
+
+        this.eventsDAO.update(eventKey, { requestedListener: true });
+      });
+    });
+  }
+
+  async saveForLater(event, room) {
+    await this.eventsDAO.setExpire(event.id, { roomId: room.id });
```
Getriax: You can update the event body here instead of calling the unnecessarily created `eventsDAO.setExpire`.

src/book/services/resources.js
```diff
@@ -29,14 +29,33 @@ class ResourcesService {
     return attendees.filter(attendee => roomEmails.indexOf(attendee.email) === -1);
   }
 
+  filterNotRooms(roomEmails, { attendees }) {
```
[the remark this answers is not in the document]
Getriax: `filterNonRooms` then?

src/book/services/resourcesManager.js
```diff
@@ -0,0 +1,20 @@
+class ResourcesManagerService {
+  constructor(resourcesService, eventsDAO) {
+    this.resourcesService = resourcesService;
+    this.eventsDAO = eventsDAO;
+  }
+
+  async handleResponse(response) {
+    if (response.isDetected === false) {
+      const event = await this.eventsDAO.getOneSet(response.uuid);
```
Getriax: What does it mean `getOneSet`? Getting a set of what? Set of events?

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,44 @@
...
+    const eventKeys = Object.keys(events);
+
+    await map(eventKeys, async (eventKey) => {
+      const event = events[eventKey];
+      const rooms = await this.getEventRoomAttendees(event);
+
+      await map(rooms, async (room) => {
+        const id = await this.saveForLater(event, room);
+
+        this.mqtt.sendMessage(room.id, { id });
```
Getriax: `room` does not have the property `id`, check https://developers.google.com/apis-explorer/#search/calendar/calendar/v3/calendar.events.get and see how `attendees` looks like.

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,44 @@
...
+    this.roomsService = roomsService;
+    this.resourcesService = resourcesService;
+  }
+
+  async prepareAndPublish(events) {
+    const eventKeys = Object.keys(events);
+
+    await map(eventKeys, async (eventKey) => {
+      const event = events[eventKey];
+      const rooms = await this.getEventRoomAttendees(event);
```
Getriax: You are receiving the filtered `attendees` here, so please don't call it `rooms`.

src/book/book.controller.js
Getriax: All right so let's suppose that case in the opposing order. 

* The event has started 9 minutes ago
* The cron starts the task, so the device is starting to listen for 10 minutes
* No one comes, so the device sends back the false
* The room is removed

The room was removed in 19 minutes since the event start instead of 10 minutes (Or whatever the timeout is).

src/book/services/resources.js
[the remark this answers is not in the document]
kozikkam: In that case `not` is more correct than `non`, which isn't a word :p

src/book/services/resourcesManager.js
Getriax: If you are not working with an event, you should not be using `eventsDAO`, it is only to access the events data, not to perform all redis transactions 🙂

kozikkam: Okay then I guess we'll have to change it's name to `eventsDTO`.

test/book/resources.service.spec.js
```diff
@@ -194,4 +193,70 @@ describe('Resources service', () => {
       .eventually
       .rejectedWith(errors.GUESTS_CANNOT_MODIFY);
   });
+
+  it('should remove one room from an event', async () => {
```
Getriax: You are checking here if one room is being removed. However, having only one room under `attendees` key means also that all the rooms are being removed. Please add one more room that should not be removed 🙂

test/factories/EventFactory.js
```diff
@@ -69,3 +69,43 @@ module.exports.shallow = ({
     requested: confirmed ? faker.random.boolean() : false,
   });
 };
+
+module.exports.processed = ({
```
Getriax: Do we need another function for this type of event? You can change the `default` function and add `attendees` there, also create one with faker if none provided 🙂

src/book/events.dto.js
```diff
@@ -19,6 +19,14 @@ class EventsDAO {
       : JSON.stringify(eventBody);
```
Getriax: I like the changed name 👍 
However, you haven't changed the class name https://github.com/netguru/bob-the-bot/blob/058a410e1ec8ace907dabfbbc009c8901679081a/src/book/events.dto.js#L9 it would be nice if you'd correct it 🙂

src/book/book.actions.js
```diff
@@ -22,7 +23,11 @@ class BookActions {
         throw new Error('Cannot confirm the event where the rooms are removed');
       }
 
-      await this.eventsDAO.update(eventId, { confirmed });
+      await this.eventsDTO.update(eventId, { confirmed });
+
+      if (confirmed === true) {
+        await this.bookService.createAwsListenersQueueJobs({ [eventId]: event });
```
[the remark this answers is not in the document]
kozikkam: The function takes object of objects, not one object. If I change it's name as you suggested, it will be misleading.

src/book/services/mqttHandler.js
```diff
@@ -0,0 +1,51 @@
...
+
+  async saveForLater(event, attendee) {
+    const id = uuid();
+    const body = {
+      roomId: attendee.email,
+      eventId: event.id,
+    };
+
+    await this.redis.set(
+      id, JSON.stringify(body), 'ex', process.env.CONFIRMED_EXPIRE_SECONDS,
```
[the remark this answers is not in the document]
kozikkam: I can't, because I want those keys to expire. Hset has poor expire support. It can only expire the whole hash, you can't set expire on a single key in a hash.

src/book/services/resourcesManager.js
```diff
@@ -0,0 +1,24 @@
+const { assertOrThrow } = require('../../../utils');
+
+class ResourcesManagerService {
+  constructor(resourcesService, redisClient) {
+    this.resourcesService = resourcesService;
+    this.redis = redisClient;
+  }
+
+  async handleResponse(response) {
+    if (response.isDetected === false) {
```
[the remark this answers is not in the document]
kozikkam: I prefer explicit statements

src/book/services/resourcesManager.js
```diff
@@ -0,0 +1,24 @@
...
+      const event = await this.redis.get(response.uuid);
```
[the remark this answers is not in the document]
kozikkam: It's being removed automatically with [expire](https://redis.io/commands/expire)

test/book/resources.service.spec.js
kozikkam: I think you should double check. `event` - input, has two attendees. `expected` - output, has 1 attendee.

test/factories/EventFactory.js
kozikkam: I will, but then every event will have to be JSON parsed, because I am using this for values retrieved from google api, not database.

PR #103, opened 2018-11-14
Integrate Rollbar with bot

krzyurb:


utils/SlackActionsMultiplexer/src/SlackActionsMultiplexer.js
```diff
@@ -73,6 +75,7 @@ class SlackActionsMultiplexer {
         this.slashActionsMultiplexer.choose(text || command);
         await this.slashActionsMultiplexer.chosen.action(res, req.body, this.robot);
       } catch (error) {
+        Rollbar.error(error);
```
Getriax: We are catching the errors in the `*.action.js` files so there won't be any error passed here.

Check it here:
https://github.com/netguru/bob-the-bot/blob/master/src/bdSupport/src/bdSupport.actions.js#L106

Could you please remove the error catching from actions so it will be logged here? 🙂

krzyurb: done

PR #105, opened 2018-11-16
Book the bot refactor

krzyurb:


src/booking/book.actions.js
```diff
@@ -1,10 +1,15 @@
 const autoBind = require('auto-bind');
 const moment = require('moment');
 
+const RoomsService = require('./rooms.service');
+const RedisClientFactory = require('./../clients/RedisClientFactory');
```
kozikkam: I think this should be done in `scripts/booking.js`, not here. We should stick to injecting dependencies from one main file, rather than doing a bit of this and a bit of that.

test/booking/rooms.service.spec.js
```diff
@@ -0,0 +1,199 @@
...
+describe('Rooms service', () => {
+  let service = null;
+
+  beforeEach(() => {
+    service = new Service(() => { });
+  });
+
+  afterEach(() => sinon.restore());
+
+  it('should return all conference rooms', async () => {
```
kozikkam: `all conference room emails`

src/booking/rooms.service.js
```diff
@@ -0,0 +1,48 @@
...
+    if (!this.rooms.length) {
+      await this.getRooms();
+    }
+
+    return attendees.filter(attendee => !this.rooms.includes(attendee.email));
+  }
+
+  async removeRooms(event) {
+    const attendees = await this.filterRooms(event);
+    Object.assign(event, { attendees });
```
kozikkam: You should never modify function arguments.

kozikkam: You should never modify function arguments.

test/booking/booking.spec.js
```diff
@@ -0,0 +1,167 @@
...
+          id: 'AEZAKMI',
+          guestsCanModify: true,
+          creator: { email: 'a@email.com' },
+          attendees: [{ email: 'room@email.com' }],
+          start: { dateTime: moment(testTime) },
+        },
+      ];
+
+      sinon.stub(service, 'getGoogleClient').returns(googleClient(events));
+      sinon.stub(redisClient, 'get').returns(false);
```
kozikkam: Redis returns event if found, and null if not found.

PR #121, opened 2018-11-27
[Improvement] GoogleClientExtractorFactory

Getriax:
## Changes

* Now using a redis client `ioredis` instead of `hubot-redis-brain`
* Fixed the bug where `refresh_token` was overwritten
* Plugins are now using the client to extract the proper library from `googleapis`


Getriax:
@krzyurb fixes done! 🙂

PR #124, opened 2018-11-28
Locale files

krzyurb:
Move bot texts from code to `.yml` files

**Usage example**

```yml
test:
  example:
    'Hello {{name}}'
  array:
  - 'first'
  - 'second'
```

```js
const locales = new Locales({ plugin: 'test' });

locales.get('test.example', { name: 'Bob' }); // Hello bob
locales.getRandom('test.example'); // e.g. first
```

**JIRA**
Tech 277


utils/Locales/src/Locales.js
```diff
@@ -0,0 +1,65 @@
+const _ = require('lodash');
+const yaml = require('js-yaml');
+const Mustache = require('mustache');
+const fs = require('fs');
+const path = require('path');
+
+class Locales {
+  constructor(options = {}) {
+    this.plugin = options.plugin;
```
Getriax: You are using only `options.plugin`, will there be more options? In my opinion `constructor(plugin)` or `constructor({ plugin })` would be better, what do you think?

PR #125, opened 2018-11-28
Bob/fixes

krzyurb:
* Fix `profile` plugin
* Limit number of showed builds in `circleci` plugin


src/circleci/src/circleci.controller.js
```diff
@@ -15,7 +15,10 @@ class CircleCiController {
     const builds = await this.CircleCiService.getBranchBuilds(username, project, branch);
     const data = this.CircleCiService.extractDataFromBuilds(builds);
 
-    const renderer = new MessageRenderer('circle/build-overview', { project, branch, data });
+    const renderer = new MessageRenderer(
+      'circle/build-overview', { project, branch, data: data.slice(0, 5) },
```
Getriax: What do you think about adding the `maxListBuilds` to `circle.config.json` and use it here? 🙂

krzyurb: done

PR #127, opened 2018-11-29
[Improve] Bob/fixes2

Getriax:
## Changes

* Added statuses to Jira plugin
* fixed a bug with the expired check which was blocking the client in googleAuthenticator
* fixed a bug with events overlapping in isFree


config/plugins/jira.config.json
```diff
@@ -13,13 +13,160 @@
...
-    "blocker":"#FF0000",
-    "critical":"#FF3232",
-    "high":"#FFD532",
-    "medium":"#0000FF",
-    "low":"#00FF00",
-    "very_low":"#E8D6C0",
-    "default":"#F7E4CC",
-    "page":"#010101"
-  }
+    "blocker": "#FF0000",
```
krzyurb: I don't like that we store that data. What if someone will remove, or add a new status on jira? Isn't it possible to check ad hoc does the selected label status exists?

Getriax: I was wondering the same thing, unfortunately, there is no endpoint for statuses as for all the projects - [source](https://community.atlassian.com/t5/Jira-questions/How-to-get-all-Jira-statuses-from-a-workflow-of-an-issue-by/qaq-p/461172)

krzyurb: And is it possible to catch the error when status, users asks for doesn't exist?

Getriax: @krzyurb check it now 🙂

src/jira/src/jira.service.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable class-methods-use-this */
-const { MessageBuilder } = require('../../../utils');
+const { MessageBuilder, assertOrThrow } = require('../../../utils');
```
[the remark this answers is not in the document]
krzyurb: @Getriax

PR #130, opened 2018-11-30
Add room remove message

kozikkam:


README.md
```diff
@@ -252,6 +252,8 @@ CALENDAR_CLIENT_ID - calendar oauth client id
 CALENDAR_CLIENT_SECRET - calendar oauth secret
 CALENDAR_OAUTH_ROOM_ID - room of a person who can grant access to bots calendar
 REDIS_URL - the url to the redis
+CONFIRMED_EXPIRE_MINUTES - minutes to hold event-room data in redis when awaiting
```
Getriax: Please include only the `CONFIRMED_EXPIRE_MINUTES` here:
https://github.com/netguru/bob-the-bot/blob/Add-Room-Remove-Message/src/booking/book.actions.js#L35

Currently, it makes a string from 11 + whatever the value is. The whole value can be under this env 🙂

src/booking/book.messages.js
```diff
@@ -47,3 +47,15 @@ module.exports.removeRoomsMessage = (event) => {
     })
     .buildMessage();
 };
+
+module.exports.removeRoomMessage = (event, roomEmail) => {
```
Getriax:
```suggestion
module.exports.removeRoomMessage = (event, roomName) => {
```

Please change that to `roomName` as you are passing the room name here not its email

kozikkam:
@Getriax

PR #138, opened 2018-12-07
Leaders ladder

krzyurb:
Create leaders ladder for profile plugin.

**Usage example**
User: `profile @user leader`
Bot: `@user's leader is @leader → @leader1 → @leader2 → @leader3`

**JIRA**
TECH-390


src/profile/test/profile.spec.js
```diff
@@ -163,12 +173,18 @@ describe('Profile test suite', () => {
...
 
-    it('should inform about custom profile field of user type', async () => {
-      await room.user.say('Josh', '@hubot profile @John leader');
-      const leader = room.robot.brain.userForId(user.profile.fields['My leader'].value);
+    it('should respond with leaders ladder', async () => {
+      await room.user.say('Josh', '@hubot profile @John leader')
+      const leader = room.robot.brain.userForId(user.profile.fields.QWERTY.value);
+      const topLeader = room.robot.brain.userForId(leader.profile.fields.QWERTY.value);
+
+      sinon.stub(client.users.profile, 'get').callsFake((params) => {
```
Getriax: This array function does not require a body, also `client.users.profile.get` is an async function so please leave it as async one 🙂

src/profile/src/profile.service.js
```diff
@@ -54,10 +59,35 @@ class ProfileService {
...
       return true;
     }
+
     const customFields = await this.getCustomFields();
     return customFields.map(field => field.label).includes(askedField);
   }
 
+  async leadersLadder(id) {
+    const { profile } = await this.client.users.profile.get({ user: id });
+    const leaderField = process.env.LEADER_FIELD_ID;
```
Getriax: Please add this environment variable to the documentation 🙂

src/profile/test/profile.spec.js
```diff
@@ -163,12 +173,18 @@ describe('Profile test suite', () => {
...
+        return room.robot.brain.userForId(params.user);
+      });
+
       const expected = [
         ['Josh', '@hubot profile @John leader'],
-        ['hubot', `@John's leader is @${leader.name}`],
+        ['hubot', `@John's leader is @${leader.name} → @${topLeader.name}`],
```
Getriax: What do you think about changing the `@John's leader is` to `@John's leaders are`? There are multiple people mentioned 🙂

krzyurb:
@Getriax fixes done

PR #144, opened 2018-12-10
Leaders plugin

krzyurb:
Move `leaders` function to separated plugin


src/leaders/src/leaders.service.js
```diff
@@ -0,0 +1,43 @@
...
+        user = user.leader;
+      }
+
+      return leaders.join(' → ');
+    }
+
+    return user.leader;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
Getriax: I think that we should disable eslint rules on the top of the file `/* eslint-disable class-methods-use-this */` it is much clearer then what rules are disabled. Could you fix that? 🙂

